PR #1, opened 2015-01-21
Ensure the proguard configuration is correct

ConradIrwin:
/cc @loopj

### DIFF
--- a/src/main/groovy/com/bugsnag/android/gradle/BugsnagPlugin.groovy
+++ b/src/main/groovy/com/bugsnag/android/gradle/BugsnagPlugin.groovy
@@ -30,8 +30,20 @@ class BugsnagPlugin implements Plugin<Project> {
                 return
             }
 
-            // Create Bugsnag post-build task
             def variantName = variant.name.capitalize()
+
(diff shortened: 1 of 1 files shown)


snmaynard:
@loopj does this seem reasonable?

What effect would this have? Would it mean we couldn't de-obfuscate proguard stack traces using the gradle plugin?

ConradIrwin:
This adds a new line of Proguard configuration which is merged with the existing project settings intelligently. We probably also want to document the required configuration on the README for this plugin.

It's not currently possible to disable this behaviour, but I can't think of a reason we'd need to.

snmaynard:
I meant more what does this fix - however I can see my question didnt actually ask that

ConradIrwin:
It fixes the ability to automatically upload the proguard file. If you don't specify `-keepattributes LineNumberTable` the upload server returns a 4xx error because we can't reverse the proguard mappings.

I documented the fact that you need to set this config on https://bugsnag.com/docs/notifiers/android/proguard, but when I came to add it to the README here discovered that it was actually fairly easy to do this automatically, so did that instead. (there's a support request that I think was caused by this).

PR #2, opened 2015-02-06
Set com.bugsnag.android.BUILD_UUID

ConradIrwin:


loopj:
Little bit of code duplication in here with the two xml sections

PR #7, opened 2015-09-08
Load Bugsnag tasks after project has been evaluated

loopj:


hzsweers:
Now that GitHub's finally decided I can interact with this repo... I haven't been able to test it personally anymore since I'm not longer at Flipboard, but the code looks good to me

PR #9, opened 2016-03-02
Fix obfuscation detection - fixes #8

### LINKED ISSUE #8: Bugsnag plugin broken with android plugin 1.5.+
I noticed that on crashes, the traces would still be obfuscated for a prod release with proguard.
So I checked my apk with aapt, and in the final apk, the meta-data is not set.

So I dug a bit deeper, read the plugin file, and saw that, you do that in https://github.com/bugsnag/bugsnag-android-gradle-plugin/blob/master/src/main/groovy/com/bugsnag/android/gradle/BugsnagPlugin.groovy

```
if (variant.getObfuscation() == null) {
    return
}
```

Problem: I use gradle android plugin 1.5.+, and plan to jump on the version 2.0.0 whenever it's ready (to benefit from all the cool features, such as code hot swapping and everything), and `variant.getObfuscation()` does not work anymore...

It looks like hockeyapp had the same issue, and here is how they solve it: https://github.com/x2on/gradle-hockeyapp-plugin/issues/74

If you can fix it as well it would be really useful, and would avoid me the hassle of having to create a custom gradle task that do a curl request to upload the mapping...

Cheers.

loopj:


loopj:
@kattrali FYI - fixed this issue and doing a release now

PR #13, opened 2016-03-15
Update gradle android plugin version from (1.0,) to 1.5.0

redwarp:
Holla, the plugin version 2.0.1 works well, but!
Actually, you specified the range in `compile 'com.android.tools.build:gradle:'` to `(1.0,)`, which basically means: use an Android plugin from 1.0 to whatever is available.

It means it will use the latest version, and the latest version is currently, in repositories, 2.1.0 something.
Problem is, this is not a stable version yet, and adds a lot of lint warnings our app is not prepared yet.

Which means that, with this version of bugsnag, the lint task use this 2.1.0 version and we get a bunch of warnings, and well... We will address these lint warnings when we start using android plugin version 2.0.0, but not yet.

I think it should be safer to use a fixed version. I propose 1.5.0, but I'm pretty sure it builds as well with version 1.0.0, like it was before.

If you can do that and publish a new aar on mavenCentral I would be so happy ^_^

Thanks in advance.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -20,7 +20,7 @@ dependencies {
 
     compile gradleApi()
     compile localGroovy()
-    compile 'com.android.tools.build:gradle:(1.0,)'
+    compile 'com.android.tools.build:gradle:1.5.0'
     compile 'org.apache.httpcomponents:httpclient:4.3.6'
     compile 'org.apache.httpcomponents:httpmime:4.3.6'
 }


kattrali:
Hi @redwarp, I definitely understand the problem of bringing in new/updated dependencies before you're ready for them. Locking to a major version makes sense, though it might be a little strange on the plugin side to bring down the allowed version for people already on 2.0+. Ideally, Android projects can specify independently which version makes sense.

Would it be possible in your project build configuration to lock the gradle version to `1.5.0` instead? I'm hesitant to make a change which might break the configuration of other users.

redwarp:
@kattrali our version is actually locked to `1.5.0`. The thing is, specifying the version of the plugin to, let's say, `1.0.0` does not prevent me from using 1.5.0 myself. I think the plugin uses the source for compiling, and then declare it as a dependency. Then later, in the project that uses the plugin, if there is a dependency conflict, it takes the higher one. So in this case, there might be a dependency conflict, and it will use `2.1.0` alpha.

But I might be mistaken of course.

kattrali:
No, you're right. It seemed counter-intuitive, but the dependency resolution just takes the higher version when there's a conflict, even alpha ones. o_O Thanks for the contribution, @redwarp. Merging.

PR #14, opened 2016-03-16
I heard you like comments

loopj:


src/main/groovy/com/bugsnag/android/gradle/BugsnagPlugin.groovy
```diff
@@ -20,41 +34,74 @@ class BugsnagPlugin implements Plugin<Project> {
...
 
                 // Create a Bugsnag task to add a build UUID to AndroidManifest.xml
-                BugsnagManifestTask manifestTask = project.tasks.create("processBugsnag${variant.name.capitalize()}Manifest", BugsnagManifestTask)
+                // This task must be called after "process${variantName}Manifest", since it
+                // requires that an AndroidManifest.xml exists in `build/intermediates`.
+                BugsnagManifestTask manifestTask = project.tasks.create("processBugsnag${variantName}Manifest", BugsnagManifestTask)
                 manifestTask.group = GROUP_NAME
                 manifestTask.manifestPath = manifestPath
-                manifestTask.dependsOn variantOutput.processManifest
+                manifestTask.mustRunAfter variantOutput.processManifest
```
kattrali: What is the difference between `dependsOn` and `mustRunAfter`?

loopj: `dependsOn` - go ahead and add this task as a dependency (adds the task to be run if it wasn't already)
`mustRunAfter` - if both of these tasks are told to execute, make sure X runs after Y (doesn't actually add a dependency)

loopj: https://docs.gradle.org/current/userguide/more_about_tasks.html#sec:adding_dependencies_to_tasks
https://docs.gradle.org/current/userguide/more_about_tasks.html#sec:ordering_tasks

kattrali: Gotcha. Looks good. :+1:

kattrali:
:tada:

PR #20, opened 2016-04-28
Fix README extension configuration

tmelz:
The Bugsnag plugin's extension options need to be set in build.gradle, not gradle.properties


loopj:
Thanks @tmelz! Well spotted.

PR #25, opened 2016-08-30
Add an option to overwrite a mapping when uploading

fcduarte:

PR #27, opened 2016-09-08
Add a retry mechanism configurable by a retryCount property

fcduarte:
@kattrali Please review it. Could you also generate a SNAPSHOT version at Bintray? Some of my team members would be able to test it and then we'll see if we're able to get rid of the `MultiEntityBuilder` error.

Thanks!


build.gradle
```diff
@@ -18,13 +18,15 @@ compileGroovy {
...
 
+def ApacheHttpComponentsVersion = '4.5.2'
+
 // Build dependencies
 dependencies {
     compile gradleApi()
     compile localGroovy()
-    compile 'com.android.tools.build:gradle:1.5.0'
-    compile 'org.apache.httpcomponents:httpclient:4.3.6'
-    compile 'org.apache.httpcomponents:httpmime:4.3.6'
```
kattrali: Per the previous discussion about the dependency versioning, I’d rather introduce new versions of httpclient separately from this feature, as it breaks the configuration of existing users. The plugin also supports gradle 1.5.0+, so I don’t see a reason to change the lower bound.

kattrali:
Can you separate the `retryCount` mechanism from the dependency upgrading so we can bring these in separately? The retry feature can be a small addition in a new minor release without changing the dependencies.

kattrali:
Got this sorted! Making a snapshot shortly.

kattrali:
Actually, cancel that - JCenter does not accept snapshot builds. I can push one to maven central if that helps, otherwise I’m drafting the 2.3.0 release in #28.

fcduarte:
Whichever works better for you, I just wanted to help on the test process to avoid any bugs like we had on #25

kattrali:
Ah, no worries then. I tried environment isolation with 4.3.6 (and ran `gradle clean` a ridiculous number of times), and it works.

PR #31, opened 2016-12-02
Add upload ndk symbols

Pezzah:
We are working towards allowing notifies from Android NDK. 

This change adds code to parse debug information from shared object files within the project, create mapping files and upload the mapping files to Bugsnag to allow de-symbolication of native stack traces.


src/main/java/com/bicirikdwarf/dwarf/Abbrev.java
```diff
@@ -0,0 +1,54 @@
+package com.bicirikdwarf.dwarf;
```
loopj: What is the license on this?

loopj: Also - can we not add bicirik-dwarf as a dependency instead?

src/main/groovy/com/bugsnag/android/gradle/BugsnagUploadAbstractTask.groovy
```diff
@@ -29,66 +24,65 @@ import org.gradle.api.tasks.TaskAction
     it is usually safe to have this be the absolute last task executed during
     a build.
 */
-class BugsnagUploadTask extends DefaultTask {
+class BugsnagUploadAbstractTask extends DefaultTask {
```
loopj: Add the `abstract` keyword here?

Pezzah:
It seems like there are two directories that we need to keep track of here:
`ProjectDir` the place where the source files are and the build folder is
`rootDir` the place where the `local.properties` file lives, which may be different if the source is in a  sub-project / package

src/main/groovy/com/bugsnag/android/gradle/BugsnagUploadNdkTask.groovy
```diff
@@ -0,0 +1,155 @@
...
+        if (objDumpPath != null) {
+
+            try {
+                File outputFile = new File(symbolPath.getAbsolutePath() + File.separator + arch + ".txt");
+
+                // Call objdump, redirecting output to the output file
+                ProcessBuilder builder = new ProcessBuilder(objDumpPath.toString(), "--disassemble", "--demangle", "--line-numbers", "--section=.text", sharedObject.toString())
+                builder.redirectOutput(outputFile)
+                Process process = builder.start()
+                while (process.alive) {
```
kattrali: Probably should use `process.waitFor()` here instead

src/main/groovy/com/bugsnag/android/gradle/BugsnagUploadNdkTask.groovy
```diff
@@ -0,0 +1,155 @@
...
+        // Get the path the version of objdump to use to get symbols
+        File objDumpPath = getObjDumpExecutable(arch)
+        if (objDumpPath != null) {
+
+            try {
+                File outputFile = new File(symbolPath.getAbsolutePath() + File.separator + arch + ".txt");
+
+                // Call objdump, redirecting output to the output file
+                ProcessBuilder builder = new ProcessBuilder(objDumpPath.toString(), "--disassemble", "--demangle", "--line-numbers", "--section=.text", sharedObject.toString())
+                builder.redirectOutput(outputFile)
```
kattrali: Drop lines starting with whitespace before writing to the file.

src/main/groovy/com/bugsnag/android/gradle/BugsnagUploadNdkTask.groovy
```diff
@@ -0,0 +1,155 @@
...
+                Process process = builder.start()
+                while (process.alive) {
+                    Thread.sleep(100)
+                }
+
+                return outputFile
```
kattrali: Before returning the file, we should check the status code to ensure everything is ok, and dump the error stream otherwise.

src/main/groovy/com/bugsnag/android/gradle/BugsnagUploadNdkTask.groovy
```diff
@@ -0,0 +1,155 @@
...
+        }
+
+        // Look for the shared objects in likely folders
+        File binariesFile = null;
+        for (File intDir : intermediatePath.listFiles()) {
+            if (intDir.name.equals("cmake")) { // CMake
+                binariesFile = intDir;
+            } else if (intDir.name.equals("binaries")) { // Experimental
+                binariesFile = intDir;
+            } // TODO: where do the build-ndk files get output?
```
kattrali: Is this TODO still pending?

src/main/groovy/com/bugsnag/android/gradle/BugsnagPlugin.groovy
```diff
@@ -58,6 +58,15 @@ class BugsnagPlugin implements Plugin<Project> {
                 // The location of the "intermediate" AndroidManifest.xml
                 def manifestPath = variantOutput.processManifest.manifestOutputFile
 
+                // Location where Proguard symbols are output
+                def symbolPath = variantOutput.processResources.textSymbolOutputDir
+                def intermediatePath = null;
+
+                // TODO: will there always be a symbol path? will it always be in this relative path to the intermediate directory?
```
kattrali: Is this TODO still pending?

src/main/groovy/com/bugsnag/android/gradle/BugsnagPlugin.groovy
```diff
@@ -73,13 +82,26 @@ class BugsnagPlugin implements Plugin<Project> {
...
+                uploadNdkTask.group = GROUP_NAME
+                uploadNdkTask.manifestPath = manifestPath
+                uploadNdkTask.applicationId = variant.applicationId
+                uploadNdkTask.intermediatePath = intermediatePath
+                uploadNdkTask.symbolPath = symbolPath
+                uploadNdkTask.variantName = variant.name
+                uploadNdkTask.projectDir = project.projectDir
+                uploadNdkTask.rootDir = project.rootDir
+                uploadNdkTask.toolchain = getCmakeToolchain(project, variant)
+                uploadNdkTask.mustRunAfter variantOutput.packageApplication
```
kattrali: Since this task only applies to some consumers, is there an `onlyIf` condition we could use here to determine if its appropriate before running?

src/main/groovy/com/bugsnag/android/gradle/BugsnagUploadNdkTask.groovy
Pezzah: changed

src/main/groovy/com/bugsnag/android/gradle/BugsnagUploadNdkTask.groovy
Pezzah: I think we need the information on the lines that start with whitespace. We could do some more processing and just keep the first and last lines in each section, but I'm not sure how much work we want to do here

src/main/groovy/com/bugsnag/android/gradle/BugsnagUploadNdkTask.groovy
Pezzah: Added code to check the status code and store the error stream

src/main/groovy/com/bugsnag/android/gradle/BugsnagUploadNdkTask.groovy
Pezzah: yes, this needs to be tested with the build-ndk system to see if it works using that (currently we just support cmake)

src/main/groovy/com/bugsnag/android/gradle/BugsnagPlugin.groovy
Pezzah: This is probably ok now

src/main/groovy/com/bugsnag/android/gradle/BugsnagUploadNdkTask.groovy
kattrali: > I'm not sure how much work we want to do here

Enough to not upload the entire disassembled library. First and last are at least significantly less and shouldn't require much more work.

src/main/groovy/com/bugsnag/android/gradle/BugsnagPlugin.groovy
Pezzah: I'm not sure what test we could run to see if ndk is enabled, especially considering the number of possible build setups

src/main/groovy/com/bugsnag/android/gradle/BugsnagUploadNdkTask.groovy
Pezzah: Added code to remove redundant address lines

src/main/groovy/com/bugsnag/android/gradle/BugsnagUploadNdkTask.groovy
Pezzah: Tested this, and added code to look for shared objects in <project>/obj/local which is where ndk-build seems to put them

src/main/groovy/com/bugsnag/android/gradle/BugsnagPlugin.groovy
Pezzah: We could add it as an option to the options, so you have to do something like

```
apply plugin: 'com.bugsnag.android.gradle'
bugsnag {
    ndk true
}
```

Pezzah:
The upload server changes in https://github.com/bugsnag/bugsnag-event-worker/pull/515 should allow for the full file to be uploaded, or the version with redundant addresses removed

src/main/groovy/com/bugsnag/android/gradle/BugsnagPlugin.groovy
Pezzah: Added a new option for this

src/main/groovy/com/bugsnag/android/gradle/BugsnagProguardConfigTask.groovy
```diff
@@ -34,6 +44,9 @@ class BugsnagProguardConfigTask extends DefaultTask {
         FileWriter fr = new FileWriter(file.path)
         fr.write(PROGUARD_CONFIG_SETTINGS)
         fr.write("\n")
+        if (useNdk) {
```
loopj: I think we can remove this conditional

PR #32, opened 2016-12-14
Fixed an issue with the apply code snippet.

Zlate87:
The original one was missing the quotes and once the quotes were added, it was resulting with:
`Error:Plugin with id 'com.bugsnag.android.gradle.BugsnagPlugin' not found.`

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,7 +58,7 @@ buildscript {
 You'll then need to "apply" the plugin by adding the following line to the top of your *Module Gradle Settings*, at `<project_dir>/<module_name>/build.gradle` (usually `<project_dir>/app/build.gradle`).
 
 ```groovy
-apply plugin: com.bugsnag.android.gradle.BugsnagPlugin
+apply plugin: 'com.bugsnag.android.gradle'
 ```
 
 


kattrali:
Thanks for the contribution, @Zlate87. Loading a plugin by either [id or class name should work](https://docs.gradle.org/current/userguide/plugins.html#sec:applying_binary_plugins), was there an issue with using the qualified class name without quotes?

Zlate87:
My script looked like this:

    apply plugin: com.bugsnag.android.gradle.BugsnagPlugin

    bugsnag {
        apiKey keys.bugSnagKey
        autoUpload false
    }

When I run it, it results in an error:

    * What went wrong:
    A problem occurred evaluating script.
    > Could not find property 'com' on project ':XXXXX'.

kattrali:
Gotcha. I think I’d run into the reverse issue previously (unable to load by id, so the instructions changed to use class reference), and as such it seems that for some versions of gradle, one or the other doesn’t work: https://github.com/bugsnag/bugsnag-android-gradle-plugin/commit/ab85a2026d4ae89224e583f733e20800429452a7

I’ll take a look to see if the id problem is resolved.

Zlate87:
If it helps, the project I had the issue on is using gradle v2.14.1.

kattrali:
This seems to work on the latest versions of 2.x and 3.x. Thanks again!

PR #34, opened 2017-02-03
[ENG-620] Add jack compiler support

Pezzah:
When using the Jack compiler the output mapping file is slightly different and should be uploaded to a different endpoint.

This change detects whether the project is being build using Jack and uses a different task to perform the upload to a different endpoint


kattrali:
Looks good! It needed a check for whether jackOptions existed as a property, but works now.

PR #36, opened 2017-03-10
Upload proguard mappings for every split

### LINKED ISSUE #35: Multiple outputs
When having multiple apks (because we have density/arch splits), bugsnag is only uploading the mappings for the `universal.apk` and not for the splitted ones.

How can we achieve this ? (I tried to do a PR with the changes but im not quite sure how do you differentiate one apk from another for the maps)

saantiaguilera:
The mappings are uploaded currently for only the first variant output.

Changes: 
- Enable mapping uploads for all variants
- Add `enabled` key to bugsnag extension, for easier manipulation of the plugin
- Increase timeout to 30s (this is because the uploads take rather long, and it was always timingout with 10s, do you want me to reset it?)
- Gradle 3.3

This closes #35


src/main/groovy/com/bugsnag/android/gradle/BugsnagUploadAbstractTask.groovy
```diff
@@ -26,7 +26,8 @@ import org.gradle.api.DefaultTask
 */
 abstract class BugsnagUploadAbstractTask extends DefaultTask {
     static final int MAX_RETRY_COUNT = 5
-    static final int TIMEOUT_MILLIS = 10000 // 10 seconds
+    // Uploading mappings sometimes the read response takes a while (45'' aprox). We are giving 1'30'' to read them
+    static final int TIMEOUT_MILLIS = 90000
```
saantiaguilera: The upload of the mappings are taking a bunch of time in the read part.. It gives me `SSOSocketTimeoutException with read time out` if I have 10s as value, what could be the issue ? (I fix it of course increasing the timeout)

fractalwrench:
Closing as this will be addressed in the following branch, using the Android Gradle 3.0.0 DSL: https://github.com/bugsnag/bugsnag-android-gradle-plugin/tree/multi-variant-upload

PR #38, opened 2017-05-06
Improve NDK object file generation performance

kattrali:
The performance of generating an object mapping file degrades for larger binary sizes, as scanning the binary can take a long time. This is a few small tweaks within the tight loop where the scan happens.


snmaynard:
@Pezzah can you look at this? @kattrali this makes a big perf difference?

Pezzah:
I've had a go at testing this with a larger file (258mb) to see what performance gains occur in this alg. I found that the existing alg took 3.3 seconds to parse the file on my machine. @kattrali's suggested change took less time, but the output file was the same size as the input file, implying that the regex was never matching. 

It looks like `matches` always tries to match the whole input string (so the `.*` at the end of the regex is required). If this is changed to use the `find` method then it just tries to find an instance within the input string. Using `find` the alg took 1.8 seconds to parse the file on my machine

Pezzah:
The change to define address matcher outside of the loop doesn't seem to affect performance

Issue #41, opened 2017-07-06
Applying plugin to 'com.android.library' project


CAMOBAP795:
### Problem
According current implementation `bugsnag` can not be applied to 'com.android.library' project
https://github.com/bugsnag/bugsnag-android-gradle-plugin/blob/master/src/main/groovy/com/bugsnag/android/gradle/BugsnagPlugin.groovy#L34

### Expected behavior
It will be nice if `bugsnag` plugin can be integrated in 'com.android.library' project

I can proceed with PR preparation if there is no objections

jkernech:
Any update on this issue ?

fractalwrench:
Released in v3.1.0

CAMOBAP795:
I just noticed that stack-trace de-obfuscation didn't work for library project.

@fractalwrench may you check this? Also is there a way to check uploaded `mapping.txt` on `https://app.bugsnag.com`?

fractalwrench:
@CAMOBAP795 can you open a separate issue with details on your gradle configuration, and plugin version? I can investigate using that information.

It's not currently possible for regular users to check uploaded mapping files.

CAMOBAP795:
@fractalwrench sure, thanks https://github.com/bugsnag/bugsnag-android-gradle-plugin/issues/106

PR #42, opened 2017-07-06
Add support for 'com.android.library' project

CAMOBAP795:
- https://github.com/bugsnag/bugsnag-android-gradle-plugin/issues/41

### DIFF
--- a/src/main/groovy/com/bugsnag/android/gradle/BugsnagPlugin.groovy
+++ b/src/main/groovy/com/bugsnag/android/gradle/BugsnagPlugin.groovy
@@ -1,11 +1,19 @@
 package com.bugsnag.android.gradle
 
 import com.android.build.gradle.AppPlugin
+import com.android.build.gradle.LibraryPlugin
 import com.android.build.gradle.api.ApplicationVariant
+import com.android.build.gradle.api.BaseVariant
+import com.android.build.gradle.api.LibraryVariant
 import com.android.build.gradle.internal.core.Toolchain
 import com.android.build.gradle.internal.dsl.BuildType
+
+import com.android.build.gradle.api.ApkVariantOutput
+import com.android.build.gradle.api.LibraryVariantOutput
+
 import org.gradle.api.Plugin
 import org.gradle.api.Project
+import org.gradle.api.DefaultTask
 /**
  * Gradle plugin to automatically upload ProGuard mapping files to Bugsnag.
  *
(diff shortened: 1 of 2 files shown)


kattrali:
Thank you for the report and the contribution, @CAMOBAP795. I'm a little unclear on the usage for this -- how do you use proguard upload tasks from a library instead of an app?

CAMOBAP795:
> how do you use proguard upload tasks from a library instead of an app?
I'm working on SDK for Android which may be used in many different applications

Currently we already have account in https://app.bugsnag.com and it works good except stack trace  'symbolication', it's expectable because we provide pro-guarded version of library for our clients

kattrali:
Gotcha! That makes sense.

CAMOBAP795:
@kattrali 

I just looked in `com.bugsnag.android.Bugsnag` and it looks like it may not work for some cases. Example:
 - in my XXX library I use `com.bugsnag.android.Bugsnag.init(context, "xxxxxxxxx")
 - my client integrate my XXX and it integrate bugsnag also for his own crash reporting
 - client's code calls `com.bugsnag.android.Bugsnag.init(context, "yyyyyyyy") 
Expected result:
 - both my reporting (as XXX library developer) and client's reporting should works
Observed result:
 - depends on who called `com.bugsnag.android.Bugsnag.init` last, it will be 'a winner in this battle'

I know that implementation of such feature isn't easy and will take valuable resources as on SDK side and on Service side also

I just wanna to know do you have any plans to implement this?

-----

Regarding my PR what is the next steps, it will be reviewed someone? how soon I may get some feedback?

kattrali:
I think it wouldn't be too difficult actually, but it would require allowing the exception handler to be chained instead of replaced in the case that its recognized to be a bugsnag handler. [Here is the relevant code](https://github.com/bugsnag/bugsnag-android/blob/master/src/main/java/com/bugsnag/android/ExceptionHandler.java#L18-L19). For your case, instead of calling `Bugsnag.init`, it may make more sense to instantiate a `Client` object directly, so a user's initialization won't override your own (given the changes described earlier are made to the exception handler behavior).

> I just wanna to know do you have any plans to implement this?
> Regarding my PR what is the next steps, it will be reviewed someone? how soon I may get some feedback?

I think its feasible and useful for similar cases where Bugsnag is being used in an SDK, just not a high priority at the moment. It'll probably be days before a review.

CAMOBAP795:
Hi @kattrali, any update about original topic?

fractalwrench:
[Add support for 'com.android.library' project](https://trello.com/c/sFJVL5KG/9-add-support-for-comandroidlibrary-project)

fractalwrench:
Hi @CAMOBAP795 thanks for the PR. This looks good but I think we'll need to find a replacement for `variantOutput.packageLibrary` as it's deprecated. I'll see if I can find a workaround for this.

CAMOBAP795:
@fractalwrench thanks, do you need some help/info from my side?

fractalwrench:
@CAMOBAP795 if you happen to know of an alternative to `variantOutput.packageLibrary` that would be helpful - otherwise I can investigate which task will work for both applications and libraries.

CAMOBAP795:
@fractalwrench sorry cannot help with this

fractalwrench:
@CAMOBAP795 no problem. I've got a feeling that it should be possible to use `build.finalizedBy` for this, so will try that out.

fractalwrench:
Closing in favour of #61, as there have been substantial changes to the plugin since this PR was opened.

PR #45, opened 2017-08-21
Address lint warnings, add missing gradle properties, autoformat

fractalwrench:
Addresses Lint warnings for Groovy by applying auto-fix suggestions. Also adds missing gradle properties to fix failed sync on checkout, and formats code consistently.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,4 +1,2 @@
 group = com.bugsnag
 version = 2.4.2
-bintray_user=your_user
-bintray_api_key=your_api_key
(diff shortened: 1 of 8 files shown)


gradle.properties
```diff
@@ -1,2 +1,4 @@
 group = com.bugsnag
 version = 2.4.2
+bintray_user=your_user
+bintray_api_key=your_api_key
```
kattrali: If these are added here, will it override `~/.gradle/gradle.properties`? I'm sure nobody wants to commit their credentials either.

fractalwrench: thanks @kattrali - I've updated the global properties file instead.

PR #48, opened 2017-09-18
Fix manual upload tasks

fractalwrench:
Exposes the task `uploadBugsnag{variantName}Mapping` allowing manual upload again. The previous merge invoked this functionality directly from within the variantOutput closure.

As part of the changes, the Tasks now hold a reference to the variantOutput. This is because the OutputFile is no longer guaranteed to be created at the time of task creation, so they are looked up when immediately required.

A side effect of this is that in my testing so far, the assemble task needs to be run, so we'd need to update the docs to reflect that. Was there a reason behind using `package` rather than `assemble`? This is the current command I'm using to upload a file:

`./gradlew assembleJavaExampleRelease :example:uploadBugsnagJavaExampleReleaseMapping`

I've also refactored the Proguard/Jack upload tasks to fit into one class.


README.md
```diff
@@ -97,7 +97,7 @@ bugsnag {
 If you disable automatic uploading, you can still run the upload task manually, with the uploadBugsnag*Variant*Mapping task:
 
 ```shell
-./gradlew clean assembleRelease packageRelease uploadBugsnagReleaseMapping
+./gradlew clean assembleRelease packageRelease ${moduleName}:uploadBugsnag${variantName}ReleaseMapping
```
kattrali: What's an example of the `moduleName` and `variantName` values? I have a test project but `uploadBugsnagReleaseMapping` works as expected.

Issue #51, opened 2017-10-08
processBugsnagDebugProguard already exists


hzsweers:
I've been seeing this error when trying the new 3.0.0 betas (using beta2 right now)

`Error:Cannot add task ':app:processBugsnagDebugProguard' as a task with that name already exists.`

fractalwrench:
@hzsweers thanks for the report. Are you using APK splits in your build.gradle by any chance?

hzsweers:
Yeah I am

PR #52, opened 2017-10-09
APK split upload support

fractalwrench:
Initial implementation of APK split support. This is achieved by finding the supported splits in the `SplitsDiscovery` task, and calculating the manifest path.

Other changes include:
- Finalising the uploadTask on `build`, so that it only runs once
- Creation of `BugsnagVariantOutputTask` to contain common functionality between tasks
- Disabling creation of upload/process tasks for debug builds

There is one breaking change to our exposed tasks, the manual upload would now look like this:
`./gradlew clean build uploadBugsnagJavaExample-releaseMapping`

Not ready to merge as we need to discuss whether there will be any API changes internally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ Settings, at `<project dir>/<module name>/build.gradle` (usually
 
 ```groovy
 plugins {
-    id "com.bugsnag.android.gradle" version "3.0.0-beta2"
+    id "com.bugsnag.android.gradle" version "3.0.0-beta3"
 }
 ```
 
@@ -50,7 +50,7 @@ Add this plugin as a dependency in your main *Project Gradle Settings*, at `<pro
 buildscript {
     dependencies {
         // Add this line to your `dependencies` section
-        classpath 'com.bugsnag:bugsnag-android-gradle-plugin:3.0.0-beta2'
+        classpath 'com.bugsnag:bugsnag-android-gradle-plugin:3.0.0-beta3'
     }
 }
 ```
@@ -101,7 +101,9 @@ If you disable automatic uploading, you can still run the upload task manually, 
 ```
 
 For example, the following would upload the release mapping file for the `x86` split of the `javaExample` productFlavor in the `example` module: 
-:example:uploadBugsnagJavaExample-x86-releaseMapping
+```shell
+./gradlew clean build :example:uploadBugsnagJavaExample-x86-releaseMapping
+```
 
 ## Automatic Proguard Config
 
--- a/README.md
+++ b/README.md
@@ -100,7 +100,8 @@ If you disable automatic uploading, you can still run the upload task manually, 
 ./gradlew clean build uploadBugsnag${variantOutputName}Mapping
 ```
 
-In this example, $moduleName is the module being built (e.g. 'app'), and $variantName is the build variant name (e.g. 'appProRelease')
+For example, the following would upload the release mapping file for the `x86` split of the `javaExample` productFlavor in the `example` module: 
+:example:uploadBugsnagJavaExample-x86-releaseMapping
 
 ## Automatic Proguard Config
 
--- a/README.md
+++ b/README.md
@@ -94,10 +94,10 @@ bugsnag {
 }
 ```
 
-If you disable automatic uploading, you can still run the upload task manually, with the uploadBugsnag*Variant*Mapping task:
+If you disable automatic uploading, you can still run the upload task manually, with the uploadBugsnag*VariantOutput*Mapping task:
 
 ```shell
-./gradlew clean assembleRelease packageRelease ${moduleName}:uploadBugsnag${variantName}Mapping
+./gradlew clean build uploadBugsnag${variantOutputName}Mapping
 ```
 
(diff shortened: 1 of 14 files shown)


src/main/groovy/com/bugsnag/android/gradle/BugsnagPlugin.groovy
```diff
@@ -36,134 +43,133 @@ class BugsnagPlugin implements Plugin<Project> {
...
-        uploadTask.mustRunAfter output.assemble
+        // only need to be run once per variant
+        setupProguardAutoConfig(project, variant)
+        setupSplitsDiscovery(project, variant)
 
-        if (project.bugsnag.autoUpload) {
-            project.tasks.findByName("package${variantName}").finalizedBy {uploadTask}
+        // need to be run for each output
+        variant.outputs.all { output ->
+            if (!output.name.toLowerCase().endsWith("debug")) {
```
kattrali: Is upload during debug something which would ever be needed/able to be enabled?

src/test/groovy/com/bugsnag/android/gradle/SmokeTest.groovy
```diff
@@ -0,0 +1,14 @@
...
+
+import org.junit.Assert
+import org.junit.Test
+
+
+class SmokeTest {
+
+    @Test
+    public void greeterPluginAddsGreetingTaskToProject() {
+        Assert.assertTrue(2 + 2 == 4)
```
kattrali: 🤔 💭

README.md
```diff
@@ -94,10 +94,10 @@ bugsnag {
...
 
-If you disable automatic uploading, you can still run the upload task manually, with the uploadBugsnag*Variant*Mapping task:
+If you disable automatic uploading, you can still run the upload task manually, with the uploadBugsnag*VariantOutput*Mapping task:
 
 ```shell
-./gradlew clean assembleRelease packageRelease ${moduleName}:uploadBugsnag${variantName}Mapping
+./gradlew clean build uploadBugsnag${variantOutputName}Mapping
 ```
 
 In this example, $moduleName is the module being built (e.g. 'app'), and $variantName is the build variant name (e.g. 'appProRelease')
```
kattrali: Might be good to show a splits example here.

src/main/groovy/com/bugsnag/android/gradle/BugsnagPlugin.groovy
fractalwrench: I suppose if proguard were enabled in a debug build, this would produce a mapping file that could be uploaded. This doesn't seem like a common case as the functionality is disabled by default, and the typical user will probably install a debuggable APK multiple times each day.

Another thing to consider if we got rid of this check is that the debug upload tasks would currently log out a warning that a mapping file couldn't be found - we could disable this behaviour for debug builds. What are your thoughts on the best approach to take?

src/test/groovy/com/bugsnag/android/gradle/SmokeTest.groovy
fractalwrench: I've updated this to add a basic test of the plugin's functionality.

src/main/groovy/com/bugsnag/android/gradle/BugsnagPlugin.groovy
kattrali: It could be behind a (default disabled) option, like `uploadDebugBuildMappings`. It sounds like it makes sense to skip but we could still support people with an unusual configuration.

README.md
fractalwrench: I've added in a splits example.

src/main/groovy/com/bugsnag/android/gradle/BugsnagPlugin.groovy
fractalwrench: Makes sense - I've added that flag into the plugin extension.

fractalwrench:
Fixes #51 and #35

src/main/groovy/com/bugsnag/android/gradle/BugsnagUploadAbstractTask.groovy
```diff
@@ -111,6 +112,13 @@ abstract class BugsnagUploadAbstractTask extends BugsnagVariantOutputTask {
         if (project.bugsnag.overwrite || System.properties['bugsnag.overwrite']) {
             mpEntity.addPart("overwrite", new StringBody("true"))
         }
+
+        project.logger.info("apiKey: ${apiKey}")
+        project.logger.info("appId: ${applicationId}")
+        project.logger.info("versionCode: ${versionCode}")
+        project.logger.info("buildUUID: ${buildUUID}")
+        project.logger.info("versionName: ${versionName}")
+        project.logger.info("overwrite: ${project.bugsnag.overwrite}")
```
kattrali: Would this be needed at a debug level, rather than info? Its quite verbose.

fractalwrench: Yes - I've updated the log level.

PR #56, opened 2017-10-16
Handle arm splits

fractalwrench:
Handles splits which contain a hyphen in their ABI filter value.

### DIFF
--- a/src/main/groovy/com/bugsnag/android/gradle/BugsnagVariantOutputTask.groovy
+++ b/src/main/groovy/com/bugsnag/android/gradle/BugsnagVariantOutputTask.groovy
@@ -33,6 +33,8 @@ class BugsnagVariantOutputTask extends DefaultTask {
             directory = guessManifestDir(directory, split)
         } else if (tokens.length == 3) {
             directory = guessManifestDir(directory, tokens[1])
+        } else if (tokens.length > 4) {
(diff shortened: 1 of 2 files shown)


src/main/groovy/com/bugsnag/android/gradle/BugsnagVariantOutputTask.groovy
```diff
@@ -28,16 +28,11 @@ class BugsnagVariantOutputTask extends DefaultTask {
...
-            } else {
-                def density = findValueForDensityFilter(split, project.ext.splitsInfo.densityFilters)
-                def abi = findValueForAbiFilter(split, project.ext.splitsInfo.abiFilters)
-                directory = findManifestDirForSplit(density, abi, directory)
-            }
+        if (tokens.length == 4) { // should be one of the following cases: 'armeabi-v7a', 'arm64-v8a'
+            def split = tokens[1] + "-" + tokens[2] // combine back into one string
+            directory = guessManifestDir(directory, split)
+        } else if (tokens.length == 3) {
+            directory = guessManifestDir(directory, tokens[1])
```
kattrali: Should probably have another case here for if there are more than 4 tokens to throw a meaningful error and print some debugging info

fractalwrench: thanks @kattrali - I've added in logging if there are more than 4 tokens.

PR #60, opened 2017-10-23
Add splitsinfo task

### LINKED ISSUE #59: Cannot get property 'splitsInfo' on extra properties extension as it does not exist
Getting this with the latest beta on AGP 3.0, seems related to the recent apk splits work?

```
Caused by: groovy.lang.MissingPropertyException: Cannot get property 'splitsInfo' on extra properties extension as it does not exist
        at org.gradle.api.internal.plugins.DefaultExtraPropertiesExtension.getProperty(DefaultExtraPropertiesExtension.java:56)
        at com.bugsnag.android.gradle.BugsnagVariantOutputTask.getManifestPath(BugsnagVariantOutputTask.groovy:37)
        at com.bugsnag.android.gradle.BugsnagManifestTask_Decorated.getManifestPath(Unknown Source)
        at com.bugsnag.android.gradle.BugsnagManifestTask_Decorated$getManifestPath.callCurrent(Unknown Source)
        at com.bugsnag.android.gradle.BugsnagManifestTask.updateManifest(BugsnagManifestTask.groovy:24)
        at org.gradle.internal.reflect.JavaMethod.invoke(JavaMethod.java:73)
```

fractalwrench:
Adds a separate task which retains information on which APK splits are available. This is more extensible than storing the values in project properties, which isn't recommended for plugins.

Tested in the following scenarios:

`./gradlew clean build && ./gradlew build`
`./gradlew build :example:uploadBugsnagKotlinExample-mdpiX86-releaseMapping`

Fixes #59

### DIFF
--- a/src/main/groovy/com/bugsnag/android/gradle/BugsnagVariantOutputTask.groovy
+++ b/src/main/groovy/com/bugsnag/android/gradle/BugsnagVariantOutputTask.groovy
@@ -57,8 +57,8 @@ class BugsnagVariantOutputTask extends DefaultTask {
         if (SPLIT_UNIVERSAL == split) {
             directory = new File(directory, SPLIT_UNIVERSAL)
         } else {
-            def density = findValueForDensityFilter(split, task.splitsInfo.densityFilters)
-            def abi = findValueForAbiFilter(split, task.splitsInfo.abiFilters)
+            def density = findValueForDensityFilter(split, task.densityFilters)
+            def abi = findValueForAbiFilter(split, task.abiFilters)
             directory = findManifestDirForSplit(density, abi, directory)
         }
         directory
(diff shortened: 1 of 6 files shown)


fractalwrench:
@kattrali in general throwing an exception would break the build. You're right that in this case it's probably better to log an error.

src/main/groovy/com/bugsnag/android/gradle/BugsnagVariantOutputTask.groovy
```diff
@@ -45,11 +47,18 @@ class BugsnagVariantOutputTask extends DefaultTask {
     }
 
     private File guessManifestDir(File directory, String split) {
+        String taskName = "bugsnagSplitsInfo${BugsnagPlugin.taskNameForVariant(variant)}"
+        def task = project.tasks.findByName(taskName)
+
+        if (task == null) {
+            throw new IllegalStateException("No task found matching ${taskName}")
```
kattrali: I think the same issue would apply here regarding throwing for Bugsnag tasks failing.

PR #61, opened 2017-10-23
Support upload of library mapping files

fractalwrench:
Adds support for uploading mapping files from a library variant. Changes include using the `BaseVariant` superclass rather than APK/Library specific classes, and altering the tasks upon which the upload tasks depend.

There have been changes in #60 which will need to be rebased onto this and adapted, so it would be good if that could be reviewed before this.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,2 +1,2 @@
 group = com.bugsnag
-version = 3.0.0-beta3
+version = 3.0.0-beta4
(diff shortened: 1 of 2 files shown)


fractalwrench:
Fixes #41

fractalwrench:
Example library project here: https://github.com/bugsnag/bugsnag-android/pull/203

fractalwrench:
@kattrali rebased and ready for review.

PR #71, opened 2017-12-13
Auto-detect ndk project

fractalwrench:
Automatically detects the presence of an NDK project by checking for the presence of an [ExternalNativeBuild](https://google.github.io/android-gradle-dsl/current/com.android.build.gradle.internal.dsl.ExternalNativeBuild.html) task. This makes `project.bugsnag.ndk` obsolete.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,2 +1,2 @@
 group = com.bugsnag
-version = 3.1.1
+version = 3.1.2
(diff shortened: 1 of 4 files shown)


fractalwrench:
@kattrali updated to respect `project.bugsnag.ndk` if it's present, and otherwise automatically detect whether this is an NDK build.

PR #87, opened 2018-02-14
Finalise uuid generation on process manifest task

fractalwrench:
Finalises UUID generation on process manifest, rather than executing task after.


kattrali:
What does this improve?

fractalwrench:
@kattrali this moves the execution order of the BugsnagManifestTask so that it completes before the `processManifest` completes. This means that changes made to the manifest such as build UUID generation are written to the file at the appropriate time. Otherwise the Manifest in the build directory and packaged APK may differ in their contents.

kattrali:
Makes sense - thanks!

PR #89, opened 2018-03-14
Add maze harness

fractalwrench:
Adds boilerplate generated by running `mazerunner init`, and moves existing example projects into the `features/fixtures` directory.


fractalwrench:
This is now ready for a final-pass review. It would be good to know if all the scenarios here and in #98 fulfil our original goals of a mazerunner test suite, or whether more are required.

PR #92, opened 2018-03-21
Install mazerunner artefact locally

fractalwrench:
Installs the plugin as a maven artefact locally, with a high version number to avoid the potential for it to be accidentally picked up from JCenter instead. The local maven repo is also cleared on each run, so the scenarios should fail if the plugin cannot build for any reason.


features/fixtures/app/gradle.properties
```diff
@@ -1,4 +1,4 @@
-plugin_version = 3.2.5
+plugin_version=9.9.9
```
kattrali: I'd go higher. Maybe over 9000.

features/scripts/clear_local_maven_repo.sh
```diff
@@ -1,3 +1,3 @@
 #!/usr/bin/env bash
 
-mv ~/.m2/repository/com/bugsnag/bugsnag-android-gradle-plugin ~/.Trash/
+mv ~/.m2/repository/com/bugsnag/bugsnag-android-gradle-plugin ~/.Trash/ || true
```
kattrali: Why not just `rm -rf` here? That would avoid having to worry about whether the file exists.

fractalwrench: I thought it would be better having this action be reversible in case something went horrendously wrong with the script/maze-runner and it deleted files it wasn't supposed to. If you're happy I can change it to `rm -rf`

fractalwrench:
Addressed review comments by increasing plugin version + deleting rather than moving

PR #93, opened 2018-03-21
Add scenario for android library

fractalwrench:
Adds a mazerunner scenario for building a default library module


features/default_lib.feature
```diff
@@ -1,5 +1,28 @@
...
+    Then I should receive 2 requests
+
+    And the request 0 is valid for the Build API
+    And the payload field "appVersion" equals "1.0" for request 0
+    And the payload field "apiKey" equals "TEST_API_KEY" for request 0
+    And the payload field "builderName" is not null for request 0
+    And the payload field "buildTool" equals "gradle-android" for request 0
+    And the payload field "appVersionCode" equals "1" for request 0
+    And the payload field "sourceControl.provider" equals "github" for request 0
+    And the payload field "sourceControl.repository" equals "https://github.com/bugsnag/bugsnag-android-gradle-plugin.git" for request 0
```
kattrali: This is the only line which _might_ fail, based on how the repository was cloned. Probably worth stubbing in a `.git/config` at the beginning of the test suite to set a constant value.

fractalwrench: I've updated the build script to temporarily alter the origin url if this is the case.

PR #98, opened 2018-03-22
Add scenario for plugin extension overrides

fractalwrench:
Adds the following scenarios where the plugin extension is configured:
 
- Overwrite parameter set to true
- autoReportBuilds set to false
- uploadDebug flag set to true in app with debug proguard maps
- Modification of default build API info


features/default_app.feature
```diff
@@ -26,4 +26,4 @@ Scenario: Single-module default app builds successfully
     And the part "versionCode" for request 1 equals "1"
     And the part "versionName" for request 1 equals "1.0"
     And the part "appId" for request 1 equals "com.bugsnag.android.example"
-
+    And the part "overwrite" for request 1 is null
```
Cawllec: This line doesn't seem to have any defined steps.

Also the Gemfile is looking for maze-runner locally at the moment.

fractalwrench:
Updated the Gemfile to use the master branch of maze-runner.

I've also addressed the missing mazerunner step in https://github.com/bugsnag/maze-runner/pull/16

fractalwrench:
Updated to use the new multipart request syntax introduced in [#16](https://github.com/bugsnag/maze-runner/pull/16)

PR #101, opened 2018-03-22
Add react native fixture

fractalwrench:
Adds a React Native app and scenario that sends a request to the build API and mapping API.

Note that the mapping API validation currently appears to be failing on MazeRunner, and the request is returned as an empty body. The mapping file uploads fine when sent to a localhost with netcat, so I suspect that our Ruby server isn't handling the rather large 1.9Mb file correctly.


fractalwrench:
Rebased to use master branch of maze-runner, and updated RN plugin version (which was altered elsewhere in a separate PR but not updated here)

PR #102, opened 2018-03-22
Add ndk app scenario

fractalwrench:
Adds an NDK app with a scenario that builds the app. Note: similar to the React Native PR, the mapping file doesn't appear to be stored in the mazerunner requests, but is received on localhost.


fractalwrench:
The mapping request doesn't appear to be stored in mazerunner which fails the scenario, I'm guessing due to the size of the payload? There's a more detailed summary in #101 

I planned to debug this issue with Alex on Monday. This is the last remaining obstacle to a full review.

features/scripts/build_ndk_app.sh
```diff
@@ -1,4 +1,18 @@
...
+BUILD_VERSION=9000.0.0-test
+
+# Build plugin
+./gradlew build -Pversion=$BUILD_VERSION
+# Install to maven local
+mvn install:install-file \
+    -Dfile=build/libs/bugsnag-android-gradle-plugin-$BUILD_VERSION.jar  \
+    -DgroupId=com.bugsnag -DartifactId=bugsnag-android-gradle-plugin \
+    -Dversion=$BUILD_VERSION \
+    -Dpackaging=jar
```
fractalwrench: @kattrali was there a reason `mvn` should be used rather than running `./gradlew install`?

fractalwrench:
This scenario is now passing after an update to mazerunner, and is ready for review.

features/scripts/build_ndk_app.sh
kattrali: Not really, in the moment I thought "how should I put this in maven local? Oh I'll use maven". Installing to maven local from gradle is more consistent.

fractalwrench:
I've updated the NDK build script so that it uses a version of the plugin installed as part of startup in `support/env.rb`. I've also updated the plugin version so that it's consistent across all the fixtures.

PR #104, opened 2018-04-09
chore(ci): Enable e2e tests

kattrali:
Made a start. Having some issues with npm and something else which doesn't seem to be in the logs on the NDK side.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,10 +8,16 @@ android:
     - build-tools-27.0.0
     - android-27
     - sys-img-armeabi-v7a-android-21
+    - extra-google-m2repository
+    - extra-android-m2repository
 
 # see https://github.com/travis-ci/travis-ci/issues/8874
 before_install:
   - yes | sdkmanager "platforms;android-27"
+  - echo y | sdkmanager "cmake;3.6.4111459"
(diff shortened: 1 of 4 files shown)


fractalwrench:
React Native now works.

The NDK is failing due to CI using an older NDK version (13) which does not support unified headers. This results in an [exception being thrown](https://github.com/bugsnag/bugsnag-android-gradle-plugin/blob/master/src/main/groovy/com/bugsnag/android/gradle/BugsnagUploadNdkTask.groovy#L246) when finding the objdump executable, and means SO maps aren't generated/uploaded. From what I remember this was necessary to get NDK uploads working with AGP 3.

The fix for CI would presumably be to update our NDK fork to a later version.

fractalwrench:
Everything passes now. I've updated the NDK scenario so that it only builds certain ABIs, and altered travis to use an NDK installed via the standard `sdkmanager` tool.

I also needed to make a small change to the plugin itself, from `process.exitValue()` to `process.waitFor()`, which blocks until the process has completed. On CI the objdump process was not completing in time for the SO mapping file to be uploaded, which I believe is a bug. I don't think that most users would run into this issue on a regular developer machine.

Issue #109, opened 2018-05-21
Gradle 4.8-rc-1 compatibility issue.


artem-zinnatullin:
Gradle recently [published 4.8-rc-1](https://github.com/gradle/gradle/releases/tag/v4.8.0-RC1), as usual I check RCes to report issues earlier

Looks like Bugsnag has some compatibility issue (likely easy to fix, but I don't want to go into details)

```java
Caused by: org.codehaus.groovy.runtime.typehandling.GroovyCastException: Cannot cast object 'org.gradle.api.internal.collections.SortedSetElementSource@6db7e212' with class 'org.gradle.api.internal.collections.SortedSetElementSource' to class 'java.util.TreeSet'
	at com.bugsnag.android.gradle.BugsnagPlugin.isJackEnabled(BugsnagPlugin.groovy:262)
	at com.bugsnag.android.gradle.BugsnagPlugin.setupMappingFileUpload(BugsnagPlugin.groovy:106)
	at com.bugsnag.android.gradle.BugsnagPlugin$_applyBugsnagToVariant_closure5.doCall(BugsnagPlugin.groovy:94)
	at com.bugsnag.android.gradle.BugsnagPlugin.applyBugsnagToVariant(BugsnagPlugin.groovy:87)
	at org.gradle.internal.metaobject.BeanDynamicObject$GroovyObjectAdapter.invokeOpaqueMethod(BeanDynamicObject.java:579)
	at org.gradle.internal.metaobject.BeanDynamicObject$MetaClassAdapter.invokeMethod(BeanDynamicObject.java:506)
	at org.gradle.internal.metaobject.BeanDynamicObject.tryInvokeMethod(BeanDynamicObject.java:191)
	at org.gradle.internal.metaobject.ConfigureDelegate.invokeMethod(ConfigureDelegate.java:78)
	at com.bugsnag.android.gradle.BugsnagPlugin$_apply_closure1$_closure11.doCall(BugsnagPlugin.groovy:42)
	at org.gradle.api.internal.ClosureBackedAction.execute(ClosureBackedAction.java:71)
	at org.gradle.util.ConfigureUtil.configureTarget(ConfigureUtil.java:155)
	at org.gradle.util.ConfigureUtil.configure(ConfigureUtil.java:106)
	at org.gradle.util.ConfigureUtil$WrappedConfigureAction.execute(ConfigureUtil.java:167)
[...]

PR #110, opened 2018-05-22
Gradle 4.8-rc-1 compat fix - use def when finding the buildType

fractalwrench:
## Goal
Gradle 4.8-rc-1 changes `project.android.buildTypes.store` to return an `Iterable` rather than a `Set`, which can lead to a runtime error within the plugin. This changeset prevents the runtime error by using `def` instead.

### Changed
Altered affected types from specific `Set` implementation to `def`.

## Tests

Mazerunner was run locally after running `./gradlew wrapper --gradle-version 4.8-rc-1` in each of the fixture projects.

### Alternative Approaches

We could use `Iterable` instead of `def`.

### Linked issues

#109 

## Review

  - [x] Final review

- [x] Consistency across platforms for structures or concepts added or modified
- [x] Consistency between the changeset and the goal stated above
- [x] Internal consistency with the rest of the library - is there any overlap between existing interfaces and any which have been added?
- [x] Usage friction - is the proposed change in usage cumbersome or complicated?
- [x] Performance and complexity - are there any cases of unexpected O(n^3) when iterating, recursing, flat mapping, etc?
- [x] Concurrency concerns - if components are accessed asynchronously, what issues will arise
- [x] Thoroughness of added tests and any missing edge cases
- [x] Idiomatic use of the language

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### Bug fixes
 
-* Gradle 4.8-rc-1 compat fix - use def when finding the buildType
+* Add compatibility with Gradle 4.8-rc1 - use def rather than TreeSet when finding the buildType
 [Jamie Lynch](https://github.com/fractalwrench) [#110](https://github.com/bugsnag/bugsnag-android-gradle-plugin/pull/110)
 
 ## 3.2.6 (2018-04-24)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## 3.2.7 (TBD)
+
+### Bug fixes
+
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+## 3.2.7 (TBD)
+
+### Bug fixes
+
+* Gradle 4.8-rc-1 compat fix - use def when finding the buildType
```
kattrali: This summary seems unclear. It should say is happening from a user perspective, e.g. "Add compatibility with Gradle 4.8-rc1"

fractalwrench:
Looks like whatever is wrong with CI is wrong on master also.

PR #114, opened 2018-06-21
fix: Update ProGuard keep rules for Breadcrumb class

fractalwrench:
## Goal

The NDK requires access to the `Breadcrumb` class as of 1.2.0, which changed location in the bugsnag-android notifier from an inner to root class. This fix
ensures that our added ProGuard rules keep the class, making it accessible via the JNI.

## Changeset

Added proguard entry to keep new class location

## Tests

Tested against NDK example project with bugsnag-android-ndk 1.2.0, a local maven installation of the plugin, and ProGuard enabled. With the old plugin version, the notifier crashes on startup, with this fix it launches and is capable of notifying. Adding the keep rule to the ProGuard file manually also works.

## Review

  - [x] Final review

### DIFF
--- a/src/main/groovy/com/bugsnag/android/gradle/BugsnagProguardConfigTask.groovy
+++ b/src/main/groovy/com/bugsnag/android/gradle/BugsnagProguardConfigTask.groovy
@@ -17,6 +17,7 @@ class BugsnagProguardConfigTask extends DefaultTask {
     static final String PROGUARD_CONFIG_SETTINGS = """\
     -keepattributes LineNumberTable,SourceFile
     -keep class com.bugsnag.android.NativeInterface { *; }
+    -keep class com.bugsnag.android.Breadcrumb { *; }
     -keep class com.bugsnag.android.Breadcrumbs { *; }
     -keep class com.bugsnag.android.Breadcrumbs\$Breadcrumb { *; }
     -keep class com.bugsnag.android.BreadcrumbType { *; }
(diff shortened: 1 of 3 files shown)


src/main/groovy/com/bugsnag/android/gradle/BugsnagProguardConfigTask.groovy
```diff
@@ -17,6 +17,7 @@ class BugsnagProguardConfigTask extends DefaultTask {
     static final String PROGUARD_CONFIG_SETTINGS = """\
     -keepattributes LineNumberTable,SourceFile
     -keep class com.bugsnag.android.NativeInterface { *; }
+    -keep class com.bugsnag.android.Breadcrumb { *; }
     -keep class com.bugsnag.android.Breadcrumbs { *; }
     -keep class com.bugsnag.android.Breadcrumbs\$Breadcrumb { *; }
```
snmaynard: I couldnt see `com.bugsnag.android.Breadcrumbs\$Breadcrumb` but maybe i'm missing something?

snmaynard: Looks like it was removed in https://github.com/bugsnag/bugsnag-android/commit/45b454465e9ef512d2fa93a3641c6d70aaba1212#diff-326654b6e94ba89f06579d82d3119049

fractalwrench: Yep. We need to keep it in the ProGuard rules for backwards compatibility with older notifiers.

PR #116, opened 2018-07-06
Add a version check for bugsnag-android dependency

fractalwrench:
## Goal

Adds a version check for the bugsnag-android dependency, and disables writing to proguard files if the version is high enough that consumer proguard rules were already added to that artefact.

## Changeset

Adds a method which collects all the dependencies for all the project's configurations, and filters these to find the version number of bugsnag-android. Disables automatic running of the Proguard config task if the version is greater than 4.6.0.

## Tests

Manual test by changing the artefact version number in a sample project.

## Review

- [x] Commented on code changes inline explain the reasoning behind the approach
  - [x] Final review


CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+
+## 3.2.8 (2018-06-21)
```
kattrali: Should be `TBD`?

Issue #121, opened 2018-08-03
Plugin not working with Android Gradle Plugin 3.3.0-alpha04


ychescale9:
The plugin seems to be broken with latest Android Gradle Plugin 3.3.0-alpha04 released today.

### Version
* Bugsnag Gradle Plugin 3.3.0
* Android Studio 3.3 Preview 4 and Android Gradle Plugin 3.3.0-alpha04

### Steps to reproduce
* Assemble the release APK: `./gradlew assembleRelease`

### Expected behavior
Gradle task finishes sucessfully.

### Observed behavior
Build fails at `app:processBugsnagProd-releaseManifest` task:

```
org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':app:processBugsnagProd-releaseManifest'.
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.executeActions(ExecuteActionsTaskExecuter.java:110)
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.execute(ExecuteActionsTaskExecuter.java:77)
        at org.gradle.api.internal.tasks.execution.OutputDirectoryCreatingTaskExecuter.execute(OutputDirectoryCreatingTaskExecuter.java:51)
        at org.gradle.api.internal.tasks.execution.SkipCachedTaskExecuter.execute(SkipCachedTaskExecuter.java:105)
        at org.gradle.api.internal.tasks.execution.SkipUpToDateTaskExecuter.execute(SkipUpToDateTaskExecuter.java:59)
        at org.gradle.api.internal.tasks.execution.ResolveTaskOutputCachingStateExecuter.execute(ResolveTaskOutputCachingStateExecuter.java:54)
        at org.gradle.api.internal.tasks.execution.ResolveBuildCacheKeyExecuter.execute(ResolveBuildCacheKeyExecuter.java:66)
        at org.gradle.api.internal.tasks.execution.ValidatingTaskExecuter.execute(ValidatingTaskExecuter.java:59)
        at org.gradle.api.internal.tasks.execution.SkipEmptySourceFilesTaskExecuter.execute(SkipEmptySourceFilesTaskExecuter.java:101)
        at org.gradle.api.internal.tasks.execution.FinalizeInputFilePropertiesTaskExecuter.execute(FinalizeInputFilePropertiesTaskExecuter.java:44)
[...]

PR #122, opened 2018-08-03
fix: support provider api introduced in gradle 4.7

fractalwrench:
## Goal

Gradle 4.7 introduced a [provider API](https://docs.gradle.org/4.7/userguide/lazy_configuration.html) for lazy evaluation of properties. The AGP subsequently changed the type of the `manifestOutputDir` property from `File` to Provider<File>. This changeset supports that API change while maintaining backward compatibility.

## Changeset

If `manifestOutputDir` is not a `File`, assume it is using the Provider API and coerce it to a file.

## Tests

Tested manually on an app using AGP 3.1.0 and another app running AGP 3.3.0.

Additionally the mazerunner scenarios were run on AGP 3.0.1, and then updated to also [run on AGP 3.3.0](https://github.com/bugsnag/bugsnag-android-gradle-plugin/tree/mazerunner-agp330). It's worth noting that the `default_lib` feature does not pass in the 3.3.0 project - this is thought to be due to a separate issue.

### Linked issues

#121

### DIFF
--- a/src/main/groovy/com/bugsnag/android/gradle/BugsnagVariantOutputTask.groovy
+++ b/src/main/groovy/com/bugsnag/android/gradle/BugsnagVariantOutputTask.groovy
@@ -4,6 +4,8 @@ import com.android.build.gradle.api.BaseVariant
 import com.android.build.gradle.api.BaseVariantOutput
 import groovy.xml.Namespace
 import org.gradle.api.DefaultTask
+import org.gradle.api.file.Directory
+import org.gradle.api.provider.Provider
 
 import java.nio.file.Paths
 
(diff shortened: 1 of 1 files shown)


src/main/groovy/com/bugsnag/android/gradle/BugsnagVariantOutputTask.groovy
```diff
@@ -4,6 +4,8 @@ import com.android.build.gradle.api.BaseVariant
 import com.android.build.gradle.api.BaseVariantOutput
 import groovy.xml.Namespace
 import org.gradle.api.DefaultTask
+import org.gradle.api.file.Directory
+import org.gradle.api.provider.Provider
```
martin308: do we support gradle versions less than 4.7? If so would there be an issue with this `import` when used with gradle < 4.7?

fractalwrench: We support lower than 4.7. The import doesn't seem to be an issue as the mazerunner scenarios pass on CI, which uses gradle <4.7.

PR #125, opened 2018-08-17
Add step to release process to trigger buildkite manually

fractalwrench:
After discussion earlier today, it was suggested that manually triggering the BuildKite build for private mazerunner scenarios would be preferable. This changeset adds a step to the release process that details how to achieve this step.


snmaynard:
We need to make the requisite changes in buildkite too

PR #126, opened 2018-08-22
Close IO streams after the plugin has finished reading and writing

fractalwrench:
## Goal

Closes IO streams after the plugin has finished reading and writing, in a `finally` block, which is guaranteed to be executed last.

## Changeset

- Altered existing use of `Closeable` values so that `close` is called in a try-finally block. 
- Added catch block where this was previously missing, that logs a warning

## Tests

Ran mazerunner scenarios.

## Review

  - [x] Final review

### DIFF
--- a/src/main/groovy/com/bugsnag/android/gradle/BugsnagManifestTask.groovy
+++ b/src/main/groovy/com/bugsnag/android/gradle/BugsnagManifestTask.groovy
@@ -58,6 +58,8 @@ class BugsnagManifestTask extends BugsnagVariantOutputTask {
                 def printer = new XmlNodePrinter(new PrintWriter(writer))
                 printer.preserveWhitespace = true
                 printer.print(xml)
+            } catch (Exception e) {
+                project.logger.warn("Failed to update manifest with Bugsnag metadata", e)
             } finally {
                 if (writer != null) {
                     writer.close()
--- a/src/main/groovy/com/bugsnag/android/gradle/BugsnagManifestTask.groovy
+++ b/src/main/groovy/com/bugsnag/android/gradle/BugsnagManifestTask.groovy
@@ -51,10 +51,18 @@ class BugsnagManifestTask extends BugsnagVariantOutputTask {
             application.appendNode('meta-data', [(ns.name): BugsnagPlugin.BUILD_UUID_TAG, (ns.value): buildUUID])
 
             // Write the manifest file
-            def writer = new FileWriter(manifestPath)
-            def printer = new XmlNodePrinter(new PrintWriter(writer))
-            printer.preserveWhitespace = true
-            printer.print(xml)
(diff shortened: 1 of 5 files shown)


src/main/groovy/com/bugsnag/android/gradle/BugsnagManifestTask.groovy
```diff
@@ -51,10 +51,20 @@ class BugsnagManifestTask extends BugsnagVariantOutputTask {
...
 
             // Write the manifest file
-            def writer = new FileWriter(manifestPath)
-            def printer = new XmlNodePrinter(new PrintWriter(writer))
-            printer.preserveWhitespace = true
-            printer.print(xml)
+            FileWriter writer = null
+
+            try {
+                writer = new FileWriter(manifestPath)
```
artem-zinnatullin: can't you use try-with-resources here?

fractalwrench: This repo currently targets Java 1.6 at the moment, so unfortunately not without some changes.

I believe the reasoning behind targeting 1.6 is that we compile the local Gradle's distribution of Groovy, which as of [Gradle 4.0](https://docs.gradle.org/4.0/release-notes.html#groovy-upgraded-to-2.4.11) is Groovy 2.4.X. This version of Groovy still supports [1.6 JVMs](http://www.groovy-lang.org/download.html#requirements).

I'll raise an internal ticket to discuss bumping our compatibility to 1.7, as I'd hope most users of Android Studio are using a modern JDK.

artem-zinnatullin: ah I see 👍

src/main/groovy/com/bugsnag/android/gradle/BugsnagProguardConfigTask.groovy
```diff
@@ -41,10 +41,17 @@ class BugsnagProguardConfigTask extends DefaultTask {
...
+        FileWriter fr = null
+
+        try {
+            fr = new FileWriter(file.path)
+            fr.write(PROGUARD_CONFIG_SETTINGS)
+            fr.write("\n")
+        } catch (Exception e) {
+            project.logger.warn("Failed to write Bugsnag ProGuard settings", e)
+        } finally {
+            fr.close()
```
kattrali: Shouldn't there be a null check here, in case the writer was not initialized?

Pezzah:
Should we also close this reader https://github.com/bugsnag/bugsnag-android-gradle-plugin/pull/126/files#diff-ff274d12fbfb73a5d6888d7dc318c167R72

fractalwrench:
Thanks, good catch! I've updated the PR to close the additional reader.

PR #127, opened 2018-08-24
[PIPE-1131] Zip shared object mapping files

Pezzah:
## Goal

Zip shared object mapping files before sending, to reduce transfer time. Also don't process the file as zipping is very effective

This coresponding PR is required to process the zipped files
https://github.com/bugsnag/bugsnag-event-worker/pull/1043

## Design

See ROAD-570 design for reasons for the approach


src/main/groovy/com/bugsnag/android/gradle/BugsnagUploadNdkTask.groovy
```diff
@@ -155,67 +153,21 @@ class BugsnagUploadNdkTask extends BugsnagMultiPartUploadTask {
...
-            return false
+        byte[] buffer = new byte[1024];
+        int len;
+        while((len=stdout.read(buffer)) != -1){
+            zipStream.write(buffer, 0, len);
         }
-
-        return true
+        zipStream.close()
+        stdout.close()
```
fractalwrench: it'd be good to close this in a finally block (there's an open PR which cleans this up generally across the plugin)

src/main/groovy/com/bugsnag/android/gradle/BugsnagUploadNdkTask.groovy
```diff
@@ -155,67 +153,21 @@ class BugsnagUploadNdkTask extends BugsnagMultiPartUploadTask {
...
-                }
-
-                line = outReader.readLine()
-            }
-
-            writer.close()
-        } catch (IOException e) {
-            project.logger.error("failed to write symbols for " + arch + ": " + e.getMessage())
-            return false
+        byte[] buffer = new byte[1024];
```
fractalwrench: BufferedOutputStream defaults to `8192`, we could probably increase this to speed up IO

src/main/groovy/com/bugsnag/android/gradle/BugsnagUploadNdkTask.groovy
Pezzah: changed

src/main/groovy/com/bugsnag/android/gradle/BugsnagUploadNdkTask.groovy
Pezzah: changed

src/main/groovy/com/bugsnag/android/gradle/BugsnagUploadNdkTask.groovy
```diff
@@ -122,7 +123,7 @@ class BugsnagUploadNdkTask extends BugsnagMultiPartUploadTask {
                     outputDir.mkdir()
                 }
 
-                File outputFile = new File(outputDir, arch + ".txt")
+                File outputFile = new File(outputDir, arch + ".gzip")
```
loopj: [Nitpick] .gz is the typical file extension for gzipped files

PR #130, opened 2018-09-07
[PLAT-1678] Change to use new objdump command 2X

Pezzah:
## Goal

Use faster version of objdump command

Requires https://github.com/bugsnag/bugsnag-event-worker/pull/1061 to parse the new file upload

## Review

- [ ] Reviewed the test cases added for completeness and possible points for discussion
- [ ] A changelog entry was added for the goal of this pull request
- [ ] Check the scope of the changeset - is everything in the diff required for the pull request?
  - [ ] Initial review of the intended approach, not yet feature complete
  - [ ] Structural review of the classes, functions, and properties modified
  - [ ] Final review

### DIFF
--- a/src/main/groovy/com/bugsnag/android/gradle/BugsnagUploadNdkTask.groovy
+++ b/src/main/groovy/com/bugsnag/android/gradle/BugsnagUploadNdkTask.groovy
@@ -152,7 +152,7 @@ class BugsnagUploadNdkTask extends BugsnagUploadAbstractTask {
                 // Call objdump, redirecting output to the output file
                 project.logger.info("${getLogPrefix()} Calling objdump process")
 
-                ProcessBuilder builder = new ProcessBuilder(objDumpPath.toString(), "--disassemble", "--demangle", "--line-numbers", "--section=.text", sharedObject.toString())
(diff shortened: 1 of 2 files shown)


src/main/groovy/com/bugsnag/android/gradle/BugsnagUploadNdkTask.groovy
```diff
@@ -152,7 +152,7 @@ class BugsnagUploadNdkTask extends BugsnagUploadAbstractTask {
                 // Call objdump, redirecting output to the output file
                 project.logger.info("${getLogPrefix()} Calling objdump process")
 
-                ProcessBuilder builder = new ProcessBuilder(objDumpPath.toString(), "--disassemble", "--demangle", "--line-numbers", "--section=.text", sharedObject.toString())
+                ProcessBuilder builder = new ProcessBuilder(objDumpPath.toString(), "-W", "-x", "--section=.debug_line", sharedObject.toString())
```
kattrali: What subset of `-W` is actually necessary (since it wraps [several options](https://sourceware.org/binutils/docs/binutils/objdump.html))? This file size could be reduced if we know specifically which components are required.

PR #137, opened 2018-12-06
Support automatic upload when building App Bundles

### LINKED ISSUE #135: Bugsnag gradle plugin not triggering after App Bundle task
### Expected behavior
*When running the bundle gradle task for my release app bundle, I expect the bugsnag gradle task to trigger and upload my proguard mapping*

### Observed behavior
*Bugsnag gradle task never runs*

### Steps to reproduce
*Run gradle task for bundle{flavor}{buildType}. In my case*

> ./gradlew clean :app:bundleProdRelease

### Version
*bugsnag-android-gradle-plugin:3.5.0*

### Additional information
*I currently have to manually call the bugsnag gradle task after my bundle gradle task finishes, when previously our APK task would automatically trigger it.*

#### Can't comment on Issues?
Some users have been unable to comment on Github issues when an [adblocker extension is enabled](https://docs.bugsnag.com/platforms/browsers/faq/#is-bugsnag-blocked-by-ad-blockers).
We recommend temporarily disabling the extension, or if that fails, contacting support@bugsnag.com.

fractalwrench:
## Goal

v3.2.0 of the Android Gradle Plugin (AGP) introduced [App Bundles](https://developer.android.com/guide/app-bundle/), a new upload format which delegates the generation of APKs to the Google Play store, and which allows dynamic delivery of feature module APKs.

The task used to generate the App Bundle does not trigger the Bugsnag plugin's mapping file upload or build API tasks automatically - this changeset fixes that.

## Changeset

The changes are best viewed by looking at the individual commits:

1. The mazerunner example app fixture was updated to use AGP v3.2.1 to enable building App Bundles. This upgrade also required updating the gradle wrapper for the fixture to 4.6 (the minimum supported), and explicitly stating that all ABIs should be built (mips, mips64, and armeabi were disabled by default in v3.2.0).

2. Mazerunner scenarios were added for several App Bundle project setups:
- A project which builds a single APK
- A project which builds two APK [productFlavors](https://developer.android.com/studio/build/build-variants#product-flavors)
- A project where autouploading is disabled

These scenarios are broadly the same as the ones used to test `assemble` tasks.

3. The method which calculates the names of `assemble` tasks was updated to also include `bundle` tasks. This in turn adds a task dependency between each `bundle` task and our upload tasks which ensures they automatically run.

For example, in our previous gradle plugin, the following tasks would be expected to trigger a mapping file upload, NDK upload, and Build API request:
```
[assemble, assembleJavaExampleRelease, assembleJavaExample, assembleRelease]
```
whereas with these changes, the following tasks would _also_ trigger this behaviour:
```
[bundle, bundleJavaExampleRelease, bundleJavaExample, bundleRelease]
```

## Tests

In addition to existing and new mazerunner scenarios, I confirmed that a crash from an APK generated from a bundle deobfuscates correctly on the dashboard. A guide on how to achieve this is available here: https://developer.android.com/guide/app-bundle/

### Linked issues
Fixes #135


src/main/groovy/com/bugsnag/android/gradle/BugsnagPlugin.groovy
```diff
@@ -195,30 +195,44 @@ class BugsnagPlugin implements Plugin<Project> {
...
+
+    /**
+     * Finds all the task names which can be used to assemble a variant, and replaces 'assemble' with the given
+     * prefix.
+     *
+     * E.g. [bundle, bundleRelease, bundleFooRelease]
+     */
+    private static Set<String> findTaskNamesForPrefix(BaseVariantOutput output, String prefix) {
         String variantName = output.name.split("-")[0].capitalize()
         String assembleTaskName = output.assemble.name
```
Pezzah: It's a shame we can't get the "bundle" task from the build variant instead of using the assemble one and replacing the names.

PR #143, opened 2018-12-26
Resolve api key from string resources

gungoren:
## Goal

When using metada like on below example upload process is broken so api returns '@string/bugsnag_api_key' not valid key.

  <meta-data
      android:name="com.bugsnag.android.API_KEY"
      android:value="@string/bugsnag_api_key" />

## Changeset

It will fix string reference from string resources.  

## Tests

It tested for flavor and flavorless projects for both type project string resource reference resolve apiKey.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 3.X.X (TBD)
+
+* Resolve apiKey from string resource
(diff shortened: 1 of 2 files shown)


fractalwrench:
Thanks for opening this PR - unfortunately I don't think we're likely to accept this as it doesn't have any tests and also logs out the API key.

PR #146, opened 2019-01-15
Address provider api warnings

fractalwrench:
## Goal

3.4-alpha of the Android Gradle Plugin has introduced new Provider APIs, which are meant to [avoid unnecessary tasks](https://developer.android.com/studio/preview/features/?utm_source=android-studio#lazy_task_config). In a project with 3.4-alpha our plugin now logs warnings about using deprecated APIs, and suggests that we should use the Provider API instead. This changeset updates our plugin to follow these recommendations.

## Design

The following locations are affected by deprecated APIs:
```
variantOutput.getProcessManifest(): com.bugsnag.android.gradle.BugsnagPlugin.setupManifestUuidTask(BugsnagPlugin.groovy:245)
variantOutput.getAssemble(): com.bugsnag.android.gradle.BugsnagPlugin.findTaskNamesForPrefix(BugsnagPlugin.groovy:229)
variantOutput.getProcessResources(): com.bugsnag.android.gradle.BugsnagUploadNdkTask.upload(BugsnagUploadNdkTask.groovy:46)
variant.getExternalNativeBuildTasks(): com.bugsnag.android.gradle.BugsnagUploadNdkTask.upload(BugsnagUploadNdkTask.groovy:60)
```

The new APIs follow this general naming convention:

```
variant.assemble // (deprecated API) eagerly accesses the assemble task
variant.assembleProvider.get() // (new API) accesses the assemble task only when required
```

Invoking the Provider API will throw an exception if the Gradle wrapper is below 5.1-RC. This can be exploited with a try-catch blocks to use the Provider API on new gradle versions, and fall back to the eager API on older versions.

## Tests

I installed a local maven artefact into the current example project, then ran `./gradlew assembleRelease`, confirming that no warnings relating to 'sdkAppExample' were logged.

I then updated the example project to the latest Canary channel, then performed the same action. An example is available on [this branch](https://github.com/bugsnag/bugsnag-android/tree/3.4-example-app). Note that due to an unrelated bug with the shared object file copy task, the bugsnag-android-ndk integration was replaced with bugsnag-android. However, this should not affect the running of the NDK upload task.

Note: the CI is also failing due to apparently unrelated issues.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,7 @@ before_install:
   - echo y | sdkmanager "platform-tools" >/dev/null
   - echo y | sdkmanager "tools" >/dev/null
   - echo y | sdkmanager "build-tools;27.0.3" >/dev/null
+  - echo y | sdkmanager 'cmake;3.6.4111459'
   - gem install bundler
   - bundle install
   - ls $ANDROID_HOME
--- a/src/main/groovy/com/bugsnag/android/gradle/BugsnagPlugin.groovy
+++ b/src/main/groovy/com/bugsnag/android/gradle/BugsnagPlugin.groovy
@@ -305,7 +305,15 @@ class BugsnagPlugin implements Plugin<Project> {
         if (variant instanceof LibraryVariant) {
             variant.getPackageLibrary().dependsOn task
         } else {
-            variant.getPackageApplication().dependsOn task
+            resolvePackageApplication(variant).dependsOn task
+        }
+    }
+
+    static def resolvePackageApplication(BaseVariant variant) {
+        try {
+            return variant.getPackageApplicationProvider().get()
+        } catch (Throwable ignored) {
+            return variant.getPackageApplication()
         }
     }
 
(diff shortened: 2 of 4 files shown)


bengourley:
Running this on the `sdkAppExample` on `bugsnag-android` I still get the following warning:

```
WARNING: API 'variant.getPackageApplication()' is obsolete and has been replaced with 'variant.getPackageApplicationProvider()'.
```

I ran the build with `-Pandroid.debug.obsoleteApi=true` and it turned up this:

```
REASON: It is currently called from the following trace:
java.lang.Thread.getStackTrace(Thread.java:1559)
com.android.build.gradle.internal.errors.DeprecationReporterImpl.reportDeprecatedApi(DeprecationReporterImpl.kt:79)
com.android.build.gradle.internal.api.ApkVariantImpl.getPackageApplication(ApkVariantImpl.java:64)
com.android.build.gradle.internal.api.ApplicationVariantImpl_Decorated.getPackageApplication(null:-1)
com.android.build.gradle.api.ApkVariant$getPackageApplication.call(null:-1)
com.bugsnag.android.gradle.BugsnagPlugin.dependTaskOnPackageTask(BugsnagPlugin.groovy:308)
[ … truncated for brevity … ]
```

fractalwrench:
Thanks @bengourley - I've fixed the remaining issue with `packageApplication` in `sdkAppExample`.

PR #150, opened 2019-03-07
Be aware of errors during mapping uploads

nimeacuerdo:
## Goal
Allow users to be aware of errors during mapping uploads and react accordingly if needed.

For us it would be helpful as we could be generating release apks where the upload of the mapping fails, leaving us with obfuscated crashes in Bugsnag's dashboard until someone manually fixes that.

This change lets users decide if they want to fail their release build if the upload of the mapping fails if they feel so.

A new `failOnUploadError` property, that defaults to **false**, has been added. 

## Tests

Manual testing of a build with no connectivity.


fractalwrench:
Thanks for the contribution @nimeacuerdo. I've reviewed the implementation and made the change of enabling `failOnUploadError` by default. I've also added some end-to-end tests in #151, which will be used to merge this functionality down.

PR #151, opened 2019-03-13
Fail on upload error

fractalwrench:
## Goal

Adds mazerunner scenarios to verify the changes within #150, which fails the build if a mapping file upload task was not successful.

## Changeset

- Includes the implementation from #150
- Sets `failOnUploadError` to true by default
- Fails the build if an API key is empty, in addition to null, during an upload task
- If a build variant does not have obfuscation enabled via `minifyEnabled` and has not applied the DexGuard plugin, the previous behaviour was to create an upload task that failed as there would be no mapping file. This task will no longer be created.

## Tests
- Added mazerunner steps which check the exit code of a script, and for running a script that is expected to exit with a non-zero exit code
- Created mazerunner scenarios where the build is expected to fail due to:
  - an empty API key
  - using the wrong endpoint
  - a missing mapping file
- Created mazerunner scenarios where the build is expected to pass due to:
  - Uploading in a default app, with an API key, mapping file, and correct endpoint specified
  - Deobfuscation being disabled, therefore the plugin would not need to upload any mapping files

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,11 @@
+## 4.X.X (TBD)
+
+Note: this version of the plugin will fail the build if a mapping file is not uploaded successfully.
+Previously if this occurred the failure would have been logged as an error and the build would have continued.
+
+* Fail on upload error
+[#151](https://github.com/bugsnag/bugsnag-android-gradle-plugin/pull/151)
+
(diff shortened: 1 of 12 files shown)


fractalwrench:
Thanks for the review comments @Cawllec - I've addressed the feedback.

PR #153, opened 2019-03-15
Add BUILD_UUID to app bundle manifests

simonbowring:
## Goal

Add a Build UUID to [app bundle](https://developer.android.com/guide/app-bundle/) manifests. 

The Build UUID currently only gets added when generating APKs directly without using app bundles.

## Changeset

- Return a path to the bundle manifest when generating an app bundle rather than the APK manifest
- Add the Build UUID to the bundle manifest rather than the merged manifest when an app bundle is being generated
- Use the Gradle task graph to determine whether an assemble task (generate APK) is being run or bundle task (generate app bundle)
- When manually uploading a mapping file, determine which manifest file to read based on which one has a Build UUID.
  - Incidentally, both a merged manifest and bundle manifest get generated for assemble and bundle tasks, but we only put the build UUID into one of them.

## Tests
Made sure that the mazerunner tests pass.

Manual tests:
- Check that an app built with the `bundleDebug` Gradle task contains the Build UUID in the manifest in the app bundle
  - Checked with AGP versions:
    - 3.2.1
    - 3.3.0
    - 3.3.2
- Check that an app built with the `assembleDebug` Gradle task contains the Build UUID in the manifest in the APK
  - Checked with AGP versions:
    - 3.0.1 (pre-app bundles)
    - 3.2.1
    - 3.3.0
    - 3.3.2
- Check that manually uploading a mapping file after building a bundle then the build UUID is read from the bundle manifest
- Check that manually uploading a mapping file after building an APK then the build UUID is read from the merged manifest

### DIFF
--- a/src/main/groovy/com/bugsnag/android/gradle/BugsnagManifestTask.groovy
+++ b/src/main/groovy/com/bugsnag/android/gradle/BugsnagManifestTask.groovy
@@ -21,52 +21,54 @@ class BugsnagManifestTask extends BugsnagVariantOutputTask {
 
     @TaskAction
     def updateManifest() {
-        def manifestPath = getManifestPath()
+        def manifestPaths = getManifestPaths()
 
-        if (!manifestPath.exists()) {
-            return
-        }
+        // Uniquely identify the build so that we can identify the proguard file.
+        def buildUUID = UUID.randomUUID().toString()
 
-        project.logger.debug("Updating manifest with build UUID: " + manifestPath)
-
-        // Parse the AndroidManifest.xml
-        def ns = new Namespace("http://schemas.android.com/apk/res/android", "android")
-        def xml = new XmlParser().parse(manifestPath)
-
-        def application = xml.application[0]
-        if (application) {
-            def metaDataTags = application['meta-data']
-
-            // remove any old BUILD_UUID_TAG elements
-            metaDataTags.findAll {
-                (it.attributes()[ns.name] == BugsnagPlugin.BUILD_UUID_TAG)
-            }.each {
-                it.parent().remove(it)
+        manifestPaths.each { manifestPath ->
+            if (!manifestPath.exists()) {
+                return
             }
 
-            // Uniquely identify the build so that we can identify the proguard file.
-            def buildUUID = UUID.randomUUID().toString()
+            project.logger.debug("Updating manifest with build UUID: " + manifestPath)
 
-            // Add the new BUILD_UUID_TAG element
-            application.appendNode('meta-data', [(ns.name): BugsnagPlugin.BUILD_UUID_TAG, (ns.value): buildUUID])
+            // Parse the AndroidManifest.xml
+            def ns = new Namespace("http://schemas.android.com/apk/res/android", "android")
+            def xml = new XmlParser().parse(manifestPath)
 
-            // Write the manifest file
-            FileWriter writer = null
+            def application = xml.application[0]
+            if (application) {
+                def metaDataTags = application['meta-data']
 
-            try {
-                writer = new FileWriter(manifestPath)
-                def printer = new XmlNodePrinter(new PrintWriter(writer))
-                printer.preserveWhitespace = true
-                printer.print(xml)
-            } catch (Exception e) {
-                project.logger.warn("Failed to update manifest with Bugsnag metadata", e)
-            } finally {
-                if (writer != null) {
-                    writer.close()
+                // remove any old BUILD_UUID_TAG elements
+                metaDataTags.findAll {
+                    (it.attributes()[ns.name] == BugsnagPlugin.BUILD_UUID_TAG)
+                }.each {
+                    it.parent().remove(it)
                 }
+
+                // Add the new BUILD_UUID_TAG element
+                application.appendNode('meta-data', [(ns.name): BugsnagPlugin.BUILD_UUID_TAG, (ns.value): buildUUID])
+
+                // Write the manifest file
+                FileWriter writer = null
+
+                try {
+                    writer = new FileWriter(manifestPath)
+                    def printer = new XmlNodePrinter(new PrintWriter(writer))
+                    printer.preserveWhitespace = true
+                    printer.print(xml)
+                } catch (Exception e) {
+                    project.logger.warn("Failed to update manifest with Bugsnag metadata", e)
+                } finally {
+                    if (writer != null) {
+                        writer.close()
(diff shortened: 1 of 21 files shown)


src/main/groovy/com/bugsnag/android/gradle/BugsnagManifestTask.groovy
```diff
@@ -75,22 +77,24 @@ class BugsnagManifestTask extends BugsnagVariantOutputTask {
...
-        def manifestPath = getManifestPath()
+        def shouldRun = false
+        def manifestPaths = getManifestPaths()
 
-        if (!manifestPath.exists()) {
-            return false
-        }
+        manifestPaths.each { manifestPath ->
+            if (!manifestPath.exists() || shouldRun) {
+                return
```
fractalwrench: This should return a boolean value

simonbowring: As the return is within a closure, that will actually move onto the next element in the `each` loop rather than exit the method. I'll switch this (and the other loops added in this PR) to a `for` loop to make it clearer.

simonbowring:
I've made a change to the approach so that either the merged manifest, the bundle manifest or both are written to depending on which tasks are in the Gradle task graph for each variant.

This has also been tested with an app running AGP 3.3.0, and one running AGP 3.0.1

Tasks left to do:
- Test with DexGuard
- Get mazerunner tests working

src/main/groovy/com/bugsnag/android/gradle/BugsnagVariantOutputTask.groovy
```diff
@@ -30,62 +31,105 @@ class BugsnagVariantOutputTask extends DefaultTask {
...
-        }
+            // If the current manifest does not contain the build ID then try the next manifest in the list (if any)
+            if (!(manifestPath == manifestPaths.last()) && !hasBuildUuid(metaDataTags, ns)) {
+                continue
+            }
+
+            // Get the Bugsnag API key
+            apiKey = getApiKey(metaDataTags, ns)
+            if (!apiKey) {
+                project.logger.warn("Could not find apiKey in '$BugsnagPlugin.API_KEY_TAG' <meta-data> tag in your AndroidManifest.xml or in your gradle config")
```
fractalwrench: I believe we've recently changed this to throw an exception in #151

src/main/groovy/com/bugsnag/android/gradle/BugsnagVariantOutputTask.groovy
```diff
@@ -30,62 +31,105 @@ class BugsnagVariantOutputTask extends DefaultTask {
...
+                project.logger.error("Failed to find manifest at ${manifestFile}")
+            } else {
+                project.logger.info("Found manifest at ${manifestFile}")
+                manifestPaths.add(manifestFile)
+            }
+        }
+
+        // Attempt to get the bundle manifest directory if required
+        if (getBundleManifest) {
+            if (!processManifest.hasProperty("bundleManifestOutputDirectory")) {
```
fractalwrench: Is it not possible to access this value at all with AGP 3.2.1? Here's an [example](https://github.com/bugsnag/bugsnag-android-gradle-plugin/blob/master/src/main/groovy/com/bugsnag/android/gradle/BugsnagPlugin.groovy#L246) of how we conditionally access a property elsewhere that is only available in later versions of the plugin.

src/main/groovy/com/bugsnag/android/gradle/BugsnagVariantOutputTask.groovy
```diff
@@ -30,62 +31,105 @@ class BugsnagVariantOutputTask extends DefaultTask {
...
+                project.logger.error("Android Gradle plugin version 3.3.0 or higher is required to read/write the bundle manifest")
+            } else {
+                directoryBundle = processManifest.bundleManifestOutputDirectory
+                File manifestFileBundle = Paths.get(directoryBundle.toString(), variantOutput.dirName, "AndroidManifest.xml").toFile()
+                if (!manifestFileBundle.exists()) {
```
fractalwrench: It'd be possible to extract a function that takes `manifestFileBundle` and `manifestFile` as a parameter, and remove some duplicated code

fractalwrench:
I believe the mazerunner tests are failing as the fixture's gradle wrapper has been updated, which seems to alter the order in which requests are sent. I don't think this is an issue in a real world scenario and is just something that needs fixing in our tests.

src/main/groovy/com/bugsnag/android/gradle/BugsnagVariantOutputTask.groovy
simonbowring: The property itself isn't available until AGP 3.3.0. I'll see if we can determine the bundle manifest directory a different way if the property is not available, perhaps it is at a consistent relative path to the merged manifest directory,

simonbowring: Actually, the property doesn't exist for AGP 3.2.1 because the bundle manifest is the same as the merged manifest. The bundle manifest got its own directory for 3.3.0. That means that the issue that this PR is trying to solve only presents itself for AGP versions >= 3.3.0. The build UUID gets added to the bundle manifest ok for AGP 3.2.1.

src/main/groovy/com/bugsnag/android/gradle/BugsnagVariantOutputTask.groovy
simonbowring: I haven't changed this block of code, only the line numbers have changed. https://github.com/bugsnag/bugsnag-android-gradle-plugin/blob/master/src/main/groovy/com/bugsnag/android/gradle/BugsnagVariantOutputTask.groovy#L74

Maybe you mean this which changed in #151?
https://github.com/bugsnag/bugsnag-android-gradle-plugin/blob/master/src/main/groovy/com/bugsnag/android/gradle/BugsnagMultiPartUploadTask.groovy#L43

simonbowring:
I've reverted the gradle and AGP version changes in the maze fixture.

fractalwrench:
I'm happy to approve this as a short-term fix with the caveat that this has been tested to work with DexGuard. We should create a ticket to address previous concerns I've raised about this general approach at a later date.

simonbowring:
I've tested this with DexGuard, deobfuscation is working.

PR #158, opened 2019-04-12
Add codenarc static analysis plugin to project

fractalwrench:
## Goal

Adding static analysis should improve the quality of new Groovy code and catch non-idiomatic usages. It will also allow us to view existing style violations and tackle technical debt.

We should add the [CodeNarc](http://codenarc.sourceforge.net/) plugin as this is most commonly used on Groovy. If we decide to migrate code over to Java/Kotlin at a later date, then we can consider also adding Lint/Checkstyle/Detekt.

## Changeset

- Applied codenarc gradle plugin to project
- Ran codenarc on CI by default
- Used the default ruleset with the exception of suppressing of `UnnecessaryGString` as this rule had ~200 violations complaining about double quotes rather than single quotes

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,6 +30,7 @@ install:
   - sudo apt-get install -y nodejs
 
 script:
+  - ./gradlew check
   - ./gradlew clean install --stacktrace
   - ./gradlew clean build --stacktrace
   - bundle exec maze-runner -c --verbose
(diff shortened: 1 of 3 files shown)


fractalwrench:
There was a discrepancy by 1 of the number of violations between mine and Roger's local machines - we suspect this is due to some difference between our environments. As we plan on reducing the number of existing violations to 0 shortly, it was decided not to spend more time investigating.

PR #160, opened 2019-04-16
Resolve pre-existing codenarc style violations

fractalwrench:
## Goal
The CodeNarc static analysis tool was added in #158 - this PR addresses pre-existing violations, where they have been simple to address. In total 16 violations have been left in the codebase with a view to addressing them as separate tickets at a later date.

## Changeset

- Reduced violation count to 16, adjusted baseline accordingly
- Added [`@CompileStatic`](http://docs.groovy-lang.org/latest/html/gapi/groovy/transform/CompileStatic.html) where changes made this possible, to gain the benefits of type-safety
- Suppressed the following overzealous checks by default:
  - InstanceOf
  - ClassJavadoc
  - ExplicitArrayListInstantiation
  - ExplicitHashSetInstantiation
  - ExplicitHashMapInstantiation
  - JavaPackageAccess
  - UnnecessaryElseStatement
  - UnnecessaryToString

Existing violations were addressed in both the main + test sourcesets. The following made up the majority of the issues:

- Use of `def` rather than an explicit type such as `String`
- Incorrect ordering of imports
- Unnecessary use of semicolon (not required in Groovy)
- Duplicate string literals
- Not using property access syntax (e.g. `foo` instead of `getFoo()`)
- Unnecessary use of return keyword (not required in Groovy)

## Tests

- Ran existing unit tests
- Ran existing mazerunner scenarios
- Ran `./gradlew assemble` in an example Android app using Gradle 5.4.0 and AGP 3.3.0 and confirmed upload of mapping files was triggered

### DIFF
--- a/src/main/groovy/com/bugsnag/android/gradle/Abi.groovy
+++ b/src/main/groovy/com/bugsnag/android/gradle/Abi.groovy
@@ -1,5 +1,6 @@
 package com.bugsnag.android.gradle
 
+@SuppressWarnings('DuplicateStringLiteral')
 enum Abi {
 
     ARMEABI(
--- a/src/main/groovy/com/bugsnag/android/gradle/Abi.groovy
+++ b/src/main/groovy/com/bugsnag/android/gradle/Abi.groovy
@@ -26,7 +26,7 @@ enum Abi {
         "x86_64",
         "x86_64",
         "x86_64-linux-android"
-    );
+    )
 
     final String abiName
     final String toolchainPrefix
--- a/src/main/groovy/com/bugsnag/android/gradle/Abi.groovy
+++ b/src/main/groovy/com/bugsnag/android/gradle/Abi.groovy
@@ -44,6 +44,6 @@ enum Abi {
                 return value
             }
         }
-        return null
+        null
     }
 }
--- a/src/main/groovy/com/bugsnag/android/gradle/Abi.groovy
+++ b/src/main/groovy/com/bugsnag/android/gradle/Abi.groovy
@@ -1,6 +1,9 @@
 package com.bugsnag.android.gradle
 
+import groovy.transform.CompileStatic
+
 @SuppressWarnings('DuplicateStringLiteral')
+@CompileStatic
 enum Abi {
 
     ARMEABI(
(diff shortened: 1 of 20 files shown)


src/main/groovy/com/bugsnag/android/gradle/Abi.groovy
```diff
@@ -1,5 +1,9 @@
 package com.bugsnag.android.gradle
 
+import groovy.transform.CompileStatic
+
+@SuppressWarnings('DuplicateStringLiteral')
```
fractalwrench: There are duplicate strings in this enum constructor, but they represent different concepts and shouldn't be grouped together into one variable.

src/main/groovy/com/bugsnag/android/gradle/BugsnagMultiPartUploadTask.groovy
```diff
@@ -97,17 +94,18 @@ abstract class BugsnagMultiPartUploadTask extends BugsnagVariantOutputTask {
...
+        String responseEntity
         try {
             HttpResponse response = httpClient.execute(httpPost)
-            statusCode = response.getStatusLine().getStatusCode()
-            responseEntity = EntityUtils.toString(response.getEntity(), "utf-8")
-        } catch (Exception e) {
+            statusCode = response.statusLine.statusCode
+            HttpEntity entity = response.entity
+            responseEntity = EntityUtils.toString(entity, "utf-8")
+        } catch (IOException | ParseException e) {
```
fractalwrench: One of the inspections was to make the caught exception less generic - I've done this where the code documents clearly what exception types can be thrown.

src/main/groovy/com/bugsnag/android/gradle/BugsnagUploadNdkTask.groovy
```diff
@@ -50,7 +52,7 @@ class BugsnagUploadNdkTask extends BugsnagMultiPartUploadTask {
             project.logger.lifecycle("Found shared object file (${arch}) ${sharedObject}")
             sharedObjectFound = true
 
-            File outputFile = createSymbolsForSharedObject(sharedObject, arch)
+            File outputFile = generateSymbolsForSharedObject(sharedObject, arch)
```
fractalwrench: `create*` indicates a factory method in groovy - this has been renamed to avoid a warning.

PR #164, opened 2019-06-03
Extract SO files from bugsnag-android artefact

fractalwrench:
## Goal

The gradle plugin should extract shared object files from both the bugsnag-android and bugsnag-android-ndk artefacts, as in the future both will contain native bugsnag code that is required by C end-users.

## Changeset

Expanded artefact filter so that files are extracted for both bugsnag-android and bugsnag-android-ndk.

## Tests
Verified that an example project requiring the NDK can be built with bugsnag-android and a local installation of this changeset.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 4.4.0 TBD
+
+* Extract SO files from bugsnag-android artifact
+[#164](https://github.com/bugsnag/bugsnag-android-gradle-plugin/pull/164)
+
 ## 4.3.1 (2019-05-31)
 
 * Calculate correct location for objdump on windows, fixing NDK symbol upload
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 4.4.0 TBD
+
+* Extract SO files from bugsnag-android artifact
```
kattrali: This is a more significant change than these notes imply, and this may need to be more verbose to convey that and ensure that anyone who upgrades bugsnag-android without bugsnag-android-gradle-plugin can see that the latest version will fix their problem.

```suggestion
This release is companion update for bugsnag-android v4.15.0, which supports detecting and reporting C/C++ crashes without a separate library. 

Extract shared object C libraries (*.so files) from bugsnag-android artifact to support using the Native API
```

\* Didn't cross-link "Native API" since the docs might move.

fractalwrench: Sounds good. We should add something to the bugsnag-android notes too when that's released, as people are more likely to read those IMO.

PR #165, opened 2019-06-10
v4.4.0

fractalwrench:


CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+## 4.4.0 (2019-06-10)
+
+This release is companion update for bugsnag-android v4.15.0, which supports detecting and reporting C/C++ crashes without a separate library. 
+
+Extract shared object C libraries (*.so files) from bugsnag-android artifact to support using the Native API
```
bengourley: This entry is a little confusing for me. I think something like the following would make more sense:

> Since `bugsnag-android` now contains native code, update shared object extraction to include libraries from there as well as `bugsnag-android-ndk`.

PR #168, opened 2019-07-12
Update plugin to search for shared object files in new artefact locations

fractalwrench:
## Goal

As part of the `bugsnag-android` modularisation SO files can now be stored in several artefacts. This adds them to the list of artefacts in our gradle plugin whose files are copied to the appropriate build directory.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## TBD
 
-Update plugin to search for shared object files in new artefact locations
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## TBD
+
+Update plugin to search for shared object files in new artefact locations
```
kattrali: This should describe the conditions where this fixes a potential issue, like the version of bugsnag-android this affects (e.g. bugsnag-android > 4.15.0)

fractalwrench: Updated the changelog entry to be a bit more specific.

PR #175, opened 2019-10-07
feat: add ability to override android:versionCode

fractalwrench:
## Goal

Allows overriding the `android:versionCode` value reported via the manifest when uploading mapping files to `upload.bugsnag.com. This can be overridden in two ways:

1. Setting `project.bugsnag.versionCode` to an integer
2. Adding `com.bugsnag.android.VERSION_CODE` as a `meta-data` element in the manifest

For further context see https://github.com/bugsnag/bugsnag-android/pull/610

## Changeset

- Adds a `versionCode` field to the `bugsnag` plugin extension, which is null by default

If `project.bugsnag.versionCode` has been set, use this value for the version code. Otherwise, search the manifest first for the value of `com.bugsnag.android.VERSION_CODE`, and then for the value of `android:versionCode`.

## Tests
- Verified default value of `versionCode` field on plugin extension in test
- Installed local artefact and verified that the `versionCode` is set correctly in 3 cases: the default; setting via the manifest; and setting via the plugin extension

### DIFF
--- a/src/main/groovy/com/bugsnag/android/gradle/BugsnagVariantOutputTask.groovy
+++ b/src/main/groovy/com/bugsnag/android/gradle/BugsnagVariantOutputTask.groovy
@@ -112,7 +112,7 @@ class BugsnagVariantOutputTask extends DefaultTask {
             }
 
             // Get the build version
-            versionCode = getVersionCode(xml, ns)
+            versionCode = getVersionCode(metaDataTags, xml, ns)
             if (versionCode == null) {
                 project.logger.warn("Could not find 'android:versionCode' value in your AndroidManifest.xml")
                 continue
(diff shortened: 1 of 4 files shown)


src/main/groovy/com/bugsnag/android/gradle/BugsnagVariantOutputTask.groovy
```diff
@@ -135,6 +135,11 @@ class BugsnagVariantOutputTask extends DefaultTask {
             apiKey = project.bugsnag.apiKey
         } else {
             apiKey = getManifestMetaData(metaDataTags, ns, BugsnagPlugin.API_KEY_TAG)
+
+            if (apiKey == null) {
+                project.logger.warn("Could not find '$BugsnagPlugin.API_KEY_TAG' " +
```
tomlongridge: There's a mix of levels that warnings are being logged at - in `readManifestFile` or in one of the `get` methods. In fact, I believe it would log twice if API weren't found?

I suggest all these logger calls be made from `readManifestFile`.

fractalwrench: Thanks - this is a good point. I've consolidated all the warnings to be logged in `readManifestFile` instead.

PR #207, opened 2020-06-26
Support for manifest processing in AGP 4.1.0-alpha04 and above

### LINKED ISSUE #194: Fails with AGP 4.1.0-alpha04
### Expected behavior

No failure!

### Observed behavior

A breaking change causes a runtime failure

### Steps to reproduce

Checkout this commit: https://github.com/ZacSweers/CatchUp/tree/e40dccf76ff213ee1046e87fdc04366864bec248

Build: `./gradlew :app:processBugsnagXxxhdpiX86_64-debugManifest`

Observe this trace

```
Execution failed for task ':app:processBugsnagXxxhdpiX86_64-debugManifest'.
> Could not get unknown property 'manifestOutputDirectory' for task ':app:processDebugManifest' of type com.android.build.gradle.tasks.ProcessApplicationManifest.
```

### Version
4.7.4

ZacSweers:
## Goal

Resolves #194 

## Design

Part of the changes here was to compile against modern gradle and AGP APIs. Both are getting much better about type safety. Along the way, I've made AGP a `compileOnly` dependency (which is fairly standard for plugins using AGP APIs anyway). This is also safe to do since new API usages are gated off (detailed below).

This creates a new, simple `BugsnagManifestTaskV2` that uses the new official API to consume a merged manifest and output a modified one with its generated build ID. This is gated by a runtime check of the AGP version, so it should be safe to use on older versions as well.

## Changeset

See above

## Tests

Want to get thoughts on the implementation before proceeding with test setup. One thing this likely merits is setting up a means of parameterized testing against multiple gradle versions

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,5 @@ gradle-app.setting
 .idea/
 local.properties
 maze_output/
+
+*.iml
--- a/build.gradle
+++ b/build.gradle
@@ -45,8 +45,8 @@ compileGroovy {
 
 // Build dependencies
 dependencies {
-    compile gradleApi()
-    compile localGroovy()
+    compileOnly gradleApi()
+    compileOnly localGroovy()
     compileOnly 'com.android.tools.build:gradle:4.1.0-beta01'
     compile 'org.apache.httpcomponents:httpclient:4.5.2'
     compile 'org.apache.httpcomponents:httpmime:4.5.2'
--- a/build.gradle
+++ b/build.gradle
@@ -5,7 +5,7 @@ buildscript {
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.1.0'
+        classpath 'com.android.tools.build:gradle:4.1.0-beta01'
     }
 }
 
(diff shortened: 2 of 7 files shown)


autonomousapps:
I think you're missing some declared dependencies, for one thing. You are using `compileOnly`, but that is for the _main_ source set. Dependencies so-declared will not appear on testCompileClasspath. You can fix that by adding those compileOnly deps to testImplementation (you want implementation so that they're also available at test runtime).

That said, when I fixed those issues, I still cannot find a class by the name `com.android.build.gradle.tasks.NdkCompile`. That package has a lot of classes in it, but none seem to be `NdkCompile`.

phillipsam:
Hey @ZacSweers thanks for submitting this PR! We'll take a look

fractalwrench:
@ZacSweers I now have some answers that will help move this PR forward:

- Writing contributions in Kotlin is fine
- The plugin's minimum supported version has been bumped to AGP 3.4.0 and Gradle 5.1.1
- The test fixture has been updated to run against the minimum supported version, and I'm in the process of creating a separate test fixture that runs against the latest & greatest

We're planning a major version bump due to dropping support for older AGP/Gradle versions, so PRs should be targeted at [this branch](https://github.com/bugsnag/bugsnag-android-gradle-plugin/tree/v5). We plan on implementing a few other improvements (configuration avoidance, removing obsolete tasks/shims, etc) before releasing any of these changes.

Hopefully that all helps! Let me know if you have any questions about the process or queries about the codebase. Thanks again for the contribution so far

ZacSweers:
That sounds good. I'll update this PR and rebase against that branch. I started down the route that @autonomousapps mentioned as well locally, but haven't pushed it up yet as I was still debugging it a bit.

ZacSweers:
Continued in #234

PR #208, opened 2020-06-30
Update test fixtures to minimum supported version

fractalwrench:
## Goal

Updates the test fixtures to the new minimum supported version (AGP 3.4.0/Gradle 5.1.1). This required the alteration of several feature files as the scenarios currently assume that requests are received in a specific order. This request order changes between AGP versions.

## Changeset

- Updated travis.yml to use recommended caching for Gradle, and to use latest versions of Android tools
- Bump bugsnag gradle plugin AGP version for test fixtures from 3.1.0 to 3.4.0, Gradle from 4.4 to 5.1.1
- Bump gradle wrapper used to build project to 6.5.1 (this is not the same as minimum supported version)
- Commented out `default_lib.feature` for now until library support is added again (broken by changes made in AGP 3.3.0)
- Skipped lint tasks when running scenarios to save execution time
- Updated `FakeVariantImpl` to implement new methods in AGP 3.4.0


.travis.yml
```diff
@@ -33,13 +30,15 @@ install:
 
 script:
   - ./gradlew check
-  - ./gradlew clean install --stacktrace
-  - ./gradlew clean build --stacktrace
+  - ./gradlew clean install -x groovyDoc -x codenarcMain -x codenarcTest --stacktrace
+  - ./gradlew clean build -x groovyDoc -x codenarcMain -x codenarcTest --stacktrace
```
fractalwrench, line 34: This avoids running unnecessary and long-running tasks as they have already been run in `./gradlew check`.

fractalwrench:
The first error is a warning from Codenarc, the static analysis tool. I believe it's missing a file required for one of the analysis modules but that this doesn't affect linting as a whole. I think the second error is a part of a mazerunner scenario where a request is made which shouldn't succeed, and we verify that zero requests are received.

PR #210, opened 2020-07-02
Remove apiKey property from bugsnag plugin extension

fractalwrench:
## Goal

Removes the deprecated apiKey property from the `BugsnagPluginExtension` class. This means that users must specify their API key in the AndroidManifest rather than the plugin extension, which led to confusion in the past.

## Tests

Relied on existing E2E test coverage. Updated `fail_on_upload.feature` to allow specifying an empty API key in one scenario.

### DIFF
--- a/src/main/groovy/com/bugsnag/android/gradle/BugsnagVariantOutputTask.groovy
+++ b/src/main/groovy/com/bugsnag/android/gradle/BugsnagVariantOutputTask.groovy
@@ -119,7 +119,7 @@ class BugsnagVariantOutputTask extends DefaultTask {
             }
 
             // Get the Bugsnag API key
-            apiKey = getApiKey(metaDataTags, ns)
+            apiKey = getManifestMetaData(metaDataTags, ns, BugsnagPlugin.API_KEY_TAG)
             if (apiKey == null) {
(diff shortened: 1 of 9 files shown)


src/main/groovy/com/bugsnag/android/gradle/BugsnagVariantOutputTask.groovy
```diff
@@ -119,7 +119,7 @@ class BugsnagVariantOutputTask extends DefaultTask {
             }
 
             // Get the Bugsnag API key
-            apiKey = getApiKey(metaDataTags, ns)
+            apiKey = getManifestMetaData(metaDataTags, ns, BugsnagPlugin.API_KEY_TAG)
             if (apiKey == null) {
                 project.logger.warn("Could not find apiKey in '$BugsnagPlugin.API_KEY_TAG' " +
                     "<meta-data> tag in your AndroidManifest.xml or in your gradle config")
```
imjoehaines, line 125: This line still mentions looking for the API key in gradle config

fractalwrench: Updated warning message to only refer to the AndroidManifest.

PR #215, opened 2020-07-03
Add test fixtures that run against AGP 4

fractalwrench:
## Goal

Implements running all scenarios on both Android Gradle Plugin 3 and 4 test fixtures.

- Updated AGP 4 test fixtures to 4.0.0 and gradle wrapper to 6.5.1 (maximum supported versions)
- Removed redundant wait steps in the build steps
- Updated scripts to build a project in the directory set by envars


features/support/env.rb
```diff
@@ -1,5 +1,18 @@
...
+  ENV["APP_FIXTURE_DIR"] = "features/fixtures/app"
+end
+
+if ENV["NDK_FIXTURE_DIR"] == nil
+  ENV["NDK_FIXTURE_DIR"] = "features/fixtures/ndkapp"
+end
+
+if ENV["RN_FIXTURE_DIR"] == nil
+  ENV["RN_FIXTURE_DIR"] = "features/fixtures/rnapp/android"
+end
```
imjoehaines, line 14: This could be shorter, if it's not too unclear:

```suggestion
ENV["APP_FIXTURE_DIR"] ||= "features/fixtures/app"
ENV["NDK_FIXTURE_DIR"] ||= "features/fixtures/ndkapp"
ENV["RN_FIXTURE_DIR"] ||= "features/fixtures/rnapp/android"
```

`||=` will only do the assignment if the current value is falsy

.gitignore
```diff
@@ -9,4 +9,5 @@ gradle-app.setting
 .idea/
 local.properties
 maze_output/
-features/fixtures/rnapp/package-lock.json
+**/package-lock.json
+**/.cxx
```
imjoehaines, line 13: I don't think the `**/` is necessary here

```suggestion
package-lock.json
.cxx
```

> A leading "\**" followed by a slash means match in all directories. For example, "**/foo" matches file or directory "foo" anywhere, the same as pattern "foo".
> https://git-scm.com/docs/gitignore

.travis.yml
```diff
@@ -49,4 +48,12 @@ jobs:
 
     # AGP 3.4.0 E2E tests
     - name: AGP 3.4.0 E2E tests
-      script: bundle exec maze-runner -c --verbose
+      env:
+        - APP_FIXTURE_DIR=features/fixtures/app_agp340
+        - RN_FIXTURE_DIR=features/fixtures/rnapp_agp340/android
+        - NDK_FIXTURE_DIR=features/fixtures/ndkapp_agp340
+      script: bundle exec maze-runner features/agp3-features features/agp-features -c --verbose
```
imjoehaines, line 55: The Ruby Maze Runner tests use tags for version specific tests; I'm not sure it makes sense for this as there's no overlap currently, but it might be worth considering in case that changes in the future

e.g. https://github.com/bugsnag/bugsnag-ruby/blob/master/features/rails_features/user_info.feature#L3
https://github.com/bugsnag/bugsnag-ruby/blob/master/.buildkite/pipeline.yml#L56

features/support/env.rb
fractalwrench: Nice :) I did not know this

.travis.yml
fractalwrench: The current approach is what I've discussed with Alex so I'll stick with this for now - tags might make sense in the future if more versions are added to the test fixtures.

PR #217, opened 2020-07-06
Convert Groovy files to Java

fractalwrench:
## Goal

Converts some Groovy files to Java where it was possible without making large changes to the existing class.

This is an intermediate step with the ultimate long-term goal to allow the conversion of parts of the project to use Kotlin.

## Changeset

- Updated single-quotes in buildscript to double quote for consistency across project
- Converted `Abi` enum and `Call` to Java class
- Split out `BugsnagPluginExtension` and `SourceControl` into separate Java classes


src/main/groovy/com/bugsnag/android/gradle/BugsnagPluginExtension.java
```diff
@@ -0,0 +1,148 @@
...
+    Boolean ndk = null;
+    String sharedObjectPath = null;
+    String projectRoot = null;
+    boolean failOnUploadError = true;
+    int requestTimeoutMs = 60000;
+
+    // release API values
+    String builderName = null;
+    Map<String, String> metadata = null;
+    Map<String, String> objdumpPaths = null;
```
imjoehaines, line 27: If these properties aren't private, do we need the getters & setters?

fractalwrench: I believe these are required for interop with Groovy. I attempted leaving these without the getters/setters previously but this resulted in a compilation error in the NDK E2E tests because the `bugsnag.ndk` property couldn't be found. Adding accessors resolved this issue.

PR #220, opened 2020-07-07
Setup support for writing plugin in Kotlin

fractalwrench:
## Goal

Adds support for writing parts of the plugin in Kotlin rather than Java/Groovy. Additionally this converts some existing Java code added in #217 to Kotlin.

## Changeset

- Added Kotlin plugin to classpath and as compile-time dependency
- Converted Java to Kotlin using IntelliJ autoconverter

## Tests

Relied on existing E2E test coverage.


src/main/kotlin/com.bugsnag.android.gradle/Abi.kt
```diff
@@ -0,0 +1,41 @@
...
+        "i686-linux-android"
+    ),
+    X86_64(
+        "x86_64",
+        "x86_64",
+        "x86_64-linux-android"
+    );
+
+    companion object {
+        fun findByName(abiName: String): Abi? {
```
imjoehaines, line 31: I'm not really familiar with Kotlin, but I think if this was [`@JvmStatic`](https://kotlinlang.org/docs/reference/java-to-kotlin-interop.html#static-methods) you could avoid having to change how `BugsnagUploadNdkTask` calls this; is that worth doing?

src/main/kotlin/com.bugsnag.android.gradle/Call.kt
```diff
@@ -0,0 +1,48 @@
+package com.bugsnag.android.gradle
+
+import org.gradle.api.Project
+
+internal abstract class Call protected constructor(// 60 seconds
```
imjoehaines, line 5: Is this `// 60 seconds` comment still relevant? I'm not sure what it refers to :s

fractalwrench: Doesn't seem to be relevant - I've removed it from the changeset.

src/main/kotlin/com.bugsnag.android.gradle/Abi.kt
fractalwrench: I tried `@JvmStatic` and it seems that Groovy is unable to find the correct method on `Abi` without the small change in `BugsnagUploadNdkTask`. As a rewrite of `BugsnagUploadNdkTask` in Kotlin is imminent I'm assuming this will be ok for now.

imjoehaines: Makes sense 👍

PR #223, opened 2020-07-09
Make upload tasks use composition rather than inheritance

fractalwrench:
## Goal

The tasks used to make requests to the Bugsnag API currently rely on inheritance of `BugsnagMultiPartUploadTask`. Refactoring this so every task extends `DefaultTask` should make it easier to support the [incremental build API](https://docs.gradle.org/current/userguide/more_about_tasks.html#sec:up_to_date_checks) by explicitly defining the task inputs and outputs.

## Changeset

Refactored the following request tasks to Java largely as-is, with modifications made where Groovy syntax is not compatible with Java. This covers the following classes:

- `BugsnagUploadProguardTask`
- `BugsnagUploadNdkTask`
- `BugsnagReleasesTask`

`BugsnagMultiPartUploadTask` has been converted from a task into a standalone class, which is instantiated when a request needs to be made.

## Tests

Relied on existing E2E test coverage.


src/main/groovy/com/bugsnag/android/gradle/BugsnagUploadProguardTask.java
```diff
@@ -0,0 +1,130 @@
...
+            FileCollection mappingFileProvider = variant.getMappingFileProvider().getOrNull();
+
+            // We will warn about not finding a mapping file later, so there's no need to warn here
+            if (mappingFileProvider == null || mappingFileProvider.isEmpty()) {
+                return null;
+            }
+            return mappingFileProvider.getSingleFile();
+        } catch (Throwable exc) {
+            return variant.getMappingFile();
+        }
```
imjoehaines, line 106: Should this live in the `AgpCompat` class now?

fractalwrench: #226 plans on removing the `AgpCompat` class in favour of declaring the compatibility shims where they are used by individual tasks

PR #226, opened 2020-07-10
Convert BugsnagPlugin to use statically typed APIs

fractalwrench:
## Goal

Converts the `BugsnagPlugin` class to Kotlin to ensure that the plugin uses public AGP APIs where possible. This changeset removes the last of the Groovy code from the bugsnag plugin so also sets up a Kotlin style checker.

This also makes a breaking change to the API by replacing `enabledBugsnag` with a global `enabled` flag. `enabledBugsnag` was applied on individual productFlavors and buildVariants, whereas `enabled` switches off everything which is more likely to be the desired behaviour. Anyone wanting to disable specific tasks should use the [Gradle task API](https://stackoverflow.com/a/47195721) which is more suitable for this purpose.

## Changeset

- Converted `BugsnagPlugin` to Kotlin
- Added [detekt](https://github.com/arturbosch/detekt) static analysis tool for Kotlin source files, removed Groovy's CodeNarc tool
- Removed remaining Groovy compatibility shims which used dynamic APIs by replacing them with Kotlin/Java equivalent instead.
- Altered `enabledBugsnag` API to be applied from plugin extension instead

## Tests
Relied on existing test coverage

TODO:

- fix enabled check for variant


src/main/kotlin/com.bugsnag.android.gradle/BugsnagPlugin.kt
```diff
@@ -0,0 +1,320 @@
...
+        return if (ndk != null) { // always respect user override
+            ndk
+        } else { // infer whether native build or not
+            val usesCmake = android.externalNativeBuild.cmake.path != null
+            val usesNdkBuild = android.externalNativeBuild.ndkBuild.path != null
+            usesCmake || usesNdkBuild
+        }
+    }
+
+    private fun setupReleasesTask(project: Project, deps: BugsnagTaskDeps, bugsnag: BugsnagPluginExtension, android: AppExtension) {
```
imjoehaines, line 142: This method doesn't seem to use `android`; can it be removed?

PR #227, opened 2020-07-13
Call register to avoid creating unnecessary tasks

fractalwrench:
## Goal

Organizes the plugin so that it's possible to call `register()` rather than `create()` when creating the `BugsnagReleasesTask`. This means that the task will be created lazily rather than eagerly and should reduce the amount of time spent in the Configuration phase.

## Changeset
- Made each bugsnag task set its group and description within the class to reduce the responsibilities of `BugsnagPlugin`
- Reduced unnecessary properties on task such as `applicationId` on `BugsnagUploadNdkTask`
- Renamed methods which setup bugsnag tasks to better reflect that tasks are registered, not created
- `TaskProvider` is returned by bugsnag task methods and `get()` is called in `variant.packageApplicationProvider.configure {}`. This ensures that tasks are only created for variants which are currently being packaged.
- Used `ApkVariant` rather than `BaseVariant` to gain access to `packageApplicationProvider` property
- Removed `BugsnagTaskDeps` in favour of specifying each task input separately (this lays the groundwork for using the incremental build API in a future PR)
- Updated manifest UUID/release task to only run on variants where JVM/NDK mapping information requires upload

## Tests

Relied on existing test coverage.

### DIFF
--- a/src/main/java/com/bugsnag/android/gradle/BugsnagVariantOutputUtils.java
+++ b/src/main/java/com/bugsnag/android/gradle/BugsnagVariantOutputUtils.java
@@ -1,7 +1,8 @@
 package com.bugsnag.android.gradle;
 
-import com.android.build.gradle.api.BaseVariant;
-import com.android.build.gradle.api.BaseVariantOutput;
+import com.android.build.gradle.api.ApkVariant;
+import com.android.build.gradle.api.ApkVariant;
+import com.android.build.gradle.api.ApkVariantOutput;
 import com.android.build.gradle.tasks.ManifestProcessorTask;
 import org.gradle.api.Project;
 import org.gradle.api.file.Directory;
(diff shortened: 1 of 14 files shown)


src/main/java/com/bugsnag/android/gradle/BugsnagVariantOutputUtils.java
```diff
@@ -1,7 +1,8 @@
 package com.bugsnag.android.gradle;
 
-import com.android.build.gradle.api.BaseVariant;
-import com.android.build.gradle.api.BaseVariantOutput;
+import com.android.build.gradle.api.ApkVariant;
+import com.android.build.gradle.api.ApkVariant;
```
imjoehaines, line 4: These two lines are the same

```suggestion
import com.android.build.gradle.api.ApkVariant;
```

src/main/kotlin/com.bugsnag.android.gradle/BugsnagPlugin.kt
```diff
@@ -42,31 +43,34 @@ class BugsnagPlugin : Plugin<Project> {
...
-                val android = project.extensions.getByType(AppExtension::class.java)
-                android.applicationVariants.all { variant ->
-                    applyBugsnagToVariant(variant, project, bugsnag, android)
-                }
-            } else {
+            if (!project.plugins.hasPlugin(AppPlugin::class.java)) {
                 throw IllegalStateException("Must apply \'com.android.application\' first!")
             }
+            if (!bugsnag.isEnabled) {
+                return@afterEvaluate
```
imjoehaines, line 50: Just so I understand what this is doing — this would be equivalent to a `continue` if this was a for loop ?

fractalwrench: This has the effect of skipping any further statements contained in the closure, i.e.: https://github.com/bugsnag/bugsnag-android-gradle-plugin/pull/227/files/f97634a605ce3259015c530ff23ce7b802f32bff#diff-ab1672558bc8b34166118b73667a1ca8R44

PR #231, opened 2020-07-15
Rename properties on bugsnag extension that control request autoUpload

fractalwrench:
## Goal

Renames properties on the plugin extension for consistency across the API surface. The old API looked like the following:

```
bugsnag {
    autoUpload = false
    autoReportBuilds = false
    ndk = false
}
```

The new API looks like this:
```
bugsnag {
    uploadJvmMappings = false
    uploadNdkMappings = false
    reportBuilds = false
}
```

### DIFF
--- a/src/main/kotlin/com.bugsnag.android.gradle/BugsnagPlugin.kt
+++ b/src/main/kotlin/com.bugsnag.android.gradle/BugsnagPlugin.kt
@@ -138,7 +138,7 @@ class BugsnagPlugin : Plugin<Project> {
         return project.tasks.register(taskName, BugsnagUploadProguardTask::class.java) {
             it.variantOutput = output
             it.variant = variant
-            addTaskToExecutionGraph(it, variant, output, project, bugsnag, bugsnag.isAutoUpload)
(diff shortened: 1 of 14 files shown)


src/main/kotlin/com.bugsnag.android.gradle/BugsnagPlugin.kt
```diff
@@ -160,7 +160,7 @@ class BugsnagPlugin : Plugin<Project> {
             it.projectDir = project.projectDir
             it.rootDir = project.rootDir
             it.sharedObjectPath = bugsnag.sharedObjectPath
-            addTaskToExecutionGraph(it, variant, output, project, bugsnag, bugsnag.isAutoUpload)
+            addTaskToExecutionGraph(it, variant, output, project, bugsnag, true)
```
imjoehaines, line 163: Is this supposed to always be true, rather than `bugsnag.isUploadJvmMappings`?

fractalwrench: This is already gated by `bugsnag.isUploadNdkMappings` in the `registerBugsnagTasksForVariant` method

PR #234, opened 2020-07-17
Support for manifest processing in AGP 4.1.0-alpha04 and above

### LINKED ISSUE #194: Fails with AGP 4.1.0-alpha04

ZacSweers:
## Goal

Resolves #194. Continuation from #207

## Design

Part of the changes here was to compile against modern gradle and AGP APIs. Both are getting much better about type safety. Along the way, I've made AGP a `compileOnly` dependency (which is fairly standard for plugins using AGP APIs anyway). This is also safe to do since new API usages are gated off (detailed below).

This creates a new, simple `BugsnagManifestUuidTaskV2` that uses the new official API to consume a merged manifest and output a modified one with its generated build ID. This is gated by a runtime check of the AGP version, so it should be safe to use on older versions as well.

## Changeset

See above

## Tests

Want to get thoughts on the implementation before proceeding with test setup. One thing this likely merits is setting up a means of parameterized testing against multiple gradle versions

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -47,6 +47,10 @@ dependencies {
     compile "org.apache.httpcomponents:httpmime:4.5.2"
     compile "org.jetbrains.kotlin:kotlin-stdlib-jdk8:1.3.72"
 
+    // For serialization of shared data
+    implementation "com.squareup.okio:okio:2.7.0"
+    implementation "com.squareup.moshi:moshi:1.9.3"
+
     testCompile "com.android.tools.build:gradle:4.0.0"
     testCompile "junit:junit:4.12"
     testCompile "org.mockito:mockito-core:2.28.2"
--- a/build.gradle
+++ b/build.gradle
@@ -5,7 +5,7 @@ buildscript {
     }
 
     dependencies {
-        classpath "com.android.tools.build:gradle:4.1.0-beta04" // compile against 4.0.0, but maintain compat to 3.4.0
+        classpath "com.android.tools.build:gradle:4.1.0-beta04" // compile against 4.1.0-beta04, but maintain compat to 3.4.0
         classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:1.3.72"
         classpath "io.gitlab.arturbosch.detekt:detekt-gradle-plugin:1.9.1"
     }
--- a/build.gradle
+++ b/build.gradle
@@ -5,7 +5,7 @@ buildscript {
     }
 
     dependencies {
-        classpath "com.android.tools.build:gradle:4.0.0" // compile against 4.0.0, but maintain compat to 3.4.0
+        classpath "com.android.tools.build:gradle:4.1.0-beta04" // compile against 4.0.0, but maintain compat to 3.4.0
         classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:1.3.72"
         classpath "io.gitlab.arturbosch.detekt:detekt-gradle-plugin:1.9.1"
     }
@@ -42,7 +42,7 @@ repositories {
 
 // Build dependencies
 dependencies {
-    compileOnly "com.android.tools.build:gradle:4.0.0"
+    compileOnly "com.android.tools.build:gradle:4.1.0-beta04"
     compile "org.apache.httpcomponents:httpclient:4.5.2"
     compile "org.apache.httpcomponents:httpmime:4.5.2"
     compile "org.jetbrains.kotlin:kotlin-stdlib-jdk8:1.3.72"
--- a/src/main/kotlin/com.bugsnag.android.gradle/BugsnagManifestUuidTask.kt
+++ b/src/main/kotlin/com.bugsnag.android.gradle/BugsnagManifestUuidTask.kt
@@ -9,7 +9,9 @@ import org.gradle.api.file.Directory
 import org.gradle.api.file.DirectoryProperty
 import org.gradle.api.file.RegularFileProperty
 import org.gradle.api.logging.Logger
+import org.gradle.api.provider.Property
 import org.gradle.api.provider.Provider
+import org.gradle.api.tasks.Input
 import org.gradle.api.tasks.OutputFile
 import org.gradle.api.tasks.TaskAction
 import java.io.File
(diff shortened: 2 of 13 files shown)


ZacSweers:
One thing I noticed is that a lot of the new tasks don't use more modern gradle property APIs. These will be really important for being friendly to gradle's future API direction, and I've sort of demonstrated a good example of how these inputs and outputs should be wired in e40508e. There's a lot more left in other tasks, but obviously out of scope for this PR

src/main/kotlin/com.bugsnag.android.gradle/BugsnagUploadProguardTask.kt
```diff
@@ -28,7 +27,7 @@ import java.nio.file.Paths
  * it is usually safe to have this be the absolute last task executed during
  * a build.
  */
-open class BugsnagUploadProguardTask : DefaultTask() {
+abstract class BugsnagUploadProguardTask : DefaultTask(), AndroidManifestInfoReceiver {
```
fractalwrench, line 30: Declaring these tasks as abstract seems to fail the build on AGP 3.4.0/Gradle 5.1. Is it necessary to declare the tasks as abstract in this way or is there a workaround that would support backwards compatibility for these versions?

>GradleException: Cannot create a proxy class for abstract class 'BugsnagManifestUuidTask'.

src/main/kotlin/com.bugsnag.android.gradle/BugsnagManifestUuidTask.kt
```diff
@@ -7,13 +7,32 @@ import org.gradle.api.DefaultTask
...
-import java.util.UUID
+
+abstract class BaseBugsnagManifestUuidTask : DefaultTask() {
+    init {
+        group = BugsnagPlugin.GROUP_NAME
+        description = "Adds a unique build UUID to AndroidManifest to link proguard mappings to crash reports"
+    }
+
+    @get:Input
+    abstract val buildUuid: Property<String>
```
fractalwrench, line 27: My current understanding is that any property annotated with input/output is used to calculate whether a task is up-to-date or not. As the `buildUuid` input is in the base class does this mean that it will always count towards the inputs of `BugsnagManifestUuidTask/BugsnagManifestUuidTaskV2`?

src/main/kotlin/com.bugsnag.android.gradle/BugsnagUploadProguardTask.kt
ZacSweers: abstract is the standard way to make gradle tasks actually, I'm surprised to see this. I suspect it's an issue with the version of gradle, a cursory check online seems to suggest that this is an issue with using too old of a jdk. Will dig more

src/main/kotlin/com.bugsnag.android.gradle/BugsnagManifestUuidTask.kt
ZacSweers: This was something Xav from the AGP team pointed out to me. We could make this up to date based on the manifest content, but we presumably want this to update every build right? It's true that the task would basically never be up to date, but I'm not sure how else to do it without adding a lot of complexity to detekt if all the other components of a build are up to date 🤔

build.gradle
```diff
@@ -5,7 +5,7 @@ buildscript {
     }
 
     dependencies {
-        classpath "com.android.tools.build:gradle:4.0.0" // compile against 4.0.0, but maintain compat to 3.4.0
+        classpath "com.android.tools.build:gradle:4.1.0-beta04" // compile against 4.0.0, but maintain compat to 3.4.0
```
digitalbuddha, line 8: comment needs adjusting
```suggestion
        classpath "com.android.tools.build:gradle:4.1.0-beta04" // compile against 4.1.0-beta04, but maintain compat to 3.4.0
```

src/main/kotlin/com.bugsnag.android.gradle/BugsnagUploadProguardTask.kt
ZacSweers: In talking with @autonomousapps, think this might be just that Gradle 5.1 doesn't support managed properties like this yet. Trying without here: d8cdeb3

src/main/kotlin/com.bugsnag.android.gradle/BugsnagManifestUuidTask.kt
ZacSweers: I think the idea would be that the plugin should be disabled on non-release builds right?

ZacSweers: Alternatively, we could make the builduuid generation a property of `BugsnagPluginExtension` to it could be set to something deterministic in debug builds and random on CI?

src/main/kotlin/com.bugsnag.android.gradle/BugsnagUploadProguardTask.kt
fractalwrench: That seems to have done the trick!

src/main/kotlin/com.bugsnag.android.gradle/BugsnagManifestUuidTask.kt
fractalwrench: I think it should be possible to make the `BugsnagManifestUuidV2Task` up-to-date by specifying only the two manifest files as inputs/outputs and generating the `buildUuid` within the task. However, I'm happy to address that myself in a separate changeset as this is getting a bit outside the original scope of this PR.

phillipsam:

PR #236, opened 2020-07-20
Update test fixtures to run against AGP 4.1.0-beta04

fractalwrench:
## Goal

Updates the latest test fixture to run against AGP 4.1.0-beta04. This changeset uses the new AGP API from 4.1.0 for altering the manifest, which was added in #234 and #240.

Note: as this changeset drops support for APK splits in AGP >=4.1.0 I'll wait for product confirmation before merging it.

## Changeset

- Updated test fixture to use AGP 4.1.0 from AGP 4.0.0
- Included changes from #234 and #240 which use the new `android.onVariants` API for updating the manifest. This is only applied on AGP 4.1.0+.
- The bugsnag manifest processing task is now registered for each `ApkVariant` on AGP 4.1.0, whereas on lower AGP versions it is registered for each `ApkVariantOutput`. This is because there does not appear to be a good API for supporting [APK splits](https://developer.android.com/studio/build/configure-apk-splits) on AGP 4.1.0 as they have been superseded by [App Bundles](https://developer.android.com/guide/app-bundle)
- Moved APK split E2E scenarios to only run against AGP 3.4.0.


fractalwrench:
@ZacSweers thought it'd be worth flagging up that the manifest changes seem to be failing when run against an example app using AGP 4.1.0-beta04. The error message seems to suggest that the input/output values haven't been set:

>org.gradle.api.InvalidUserDataException: No value has been specified for property 'outputManifest

I'll keep the changes from #234 merged for now and think any fix can be applied using this PR. If you're still happy to contribute, would you have any idea what might be causing this behaviour? If not I can take the time to look at this myself.

ZacSweers:
Yep just that I forgot to set the output path in my previous PR :). We should just set a path in `build/intermediates/bugsnag/...` like the other one. I can make a PR and target this branch if that works?

ZacSweers:
This patch should work

```patch
Index: src/main/kotlin/com.bugsnag.android.gradle/BugsnagPlugin.kt
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- src/main/kotlin/com.bugsnag.android.gradle/BugsnagPlugin.kt	(revision 6de8a1a0395d22280d4d4fb5e5c552423bd5ef5a)
+++ src/main/kotlin/com.bugsnag.android.gradle/BugsnagPlugin.kt	(date 1595223611000)
@@ -122,9 +122,11 @@
         val buildUuidProvider = project.provider { UUID.randomUUID().toString() }
         val manifestInfoOutputFile = project.layout.buildDirectory.file("intermediates/bugsnag/manifestInfoFor${taskNameForOutput(output)}.json")
         return if (BugsnagManifestUuidTaskV2.isApplicable()) {
+            val processedManifestOutput = project.layout.buildDirectory.file("intermediates/bugsnag/processed${taskNameForOutput(output)}Manifest.xml")
             val manifestUpdater = project.tasks.register(taskName, BugsnagManifestUuidTaskV2::class.java) {
                 it.buildUuid.set(buildUuidProvider)
                 it.manifestInfoProvider.set(manifestInfoOutputFile)
+                it.outputManifest.set(processedManifestOutput)
             }
             val android = project.extensions.getByType(BaseAppModuleExtension::class.java)
             android.onVariants.withName(variant.name) {

```

fractalwrench:
>I can make a PR and target this branch if that works?

That would be great! It's worth noting that `inputManifest` is also not set on the task so the patch doesn't seem to work as-is.

ZacSweers:
hmm, inputManifest should be getting set by AGP for you there. What's the full error?

fractalwrench:
CI comment has now been addressed in #246

PR #240, opened 2020-07-23
Fix missing outputfile configuration

ZacSweers:
Ref: https://github.com/bugsnag/bugsnag-android-gradle-plugin/pull/236#issuecomment-661768050


ZacSweers:
This seems to be an issue in AGP, filed https://issuetracker.google.com/issues/161920328

fractalwrench:
Thanks for digging down into this Zac - it looks like we'll need to wait for the resolution of the issue in AGP.

ZacSweers:
it looks like it's never getting a callback in `onVariants` 🤔

ZacSweers:
So the wiring works now, but it looks like there's unrelated stuff failing (version code, mapping files, etc), possibly on the branch itself?

fractalwrench:
I made a couple of changes to get this passing:

1. The test fixture that runs against AGP 3.4.0 was failing on some of the APK splits scenarios as the task name used the name of the `ApkVariant`, rather than the `ApkVariantOutput`, which isn't unique.
2. Some of the APK split tests were failing on AGP 4.1.0, so I've moved them to only run on AGP 3.4.0.

I'm assuming that there isn't an AGP 4.1.0 API for accessing APK splits because using App Bundles is preferred. If this isn't the case and it's possible to support APK splits then I imagine further changes will be needed to support that.

Thanks again for the contribution! I'll merge this down to the `v5-fixture-dep-bump` branch and wait to confirm that there isn't an API for supporting APK splits in AGP 4.1.0 (and also alert our product folks about this).

PR #246, opened 2020-07-28
Increase test matrix for E2E scenarios

fractalwrench:
## Goal

Increases the test matrix for E2E scenarios as requested in #236. This updates the E2E scenarios to run against the following:

- AGP 3.4.0, gradle 5.1.1
- AGP 4.0.0, gradle 6.1.1
- AGP 4.1.0, gradle 6.5.1

This changeset also alters the E2E scenarios so they all run from one test fixture, and so that scenarios are skipped according to cucumber tags. This should make it much easier to add new AGP versions to the matrix in a future PR.

## Changeset

- Increases test matrix to run against 3 AGP/Gradle versions, set from the minimum requirements on the [gradle plugin release notes](https://developer.android.com/studio/releases/gradle-plugin)
- Skipped APK split scenarios on AGP 4.1.0 and higher, skipped AGP 3 scenarios on AGP 4 and higher through use of cucumber tags
- Moved all scenarios into `features` directory
- Deleted the AGP 3.4.0 specific test fixtures in favour of using one test fixture for each the JVM, NDK, and RN fixtures
- Used an `AGP_VERSION` and `GRADLE_WRAPPER_VERSION` envar to set the AGP plugin version and Gradle Wrapper version in each test fixture
- Conditionally applied bugsnag gradle plugin behind `UPDATING_GRADLEW ` envar, as the gradle wrapper needs to be updated for each fixture to the targeted version. At this point in time the bugsnag plugin is not built and this would therefore fail the build


fractalwrench:
Updated travis.yml to add in AGP 3.5.0 and AGP 3.6.0 by default too, as this took little effort to get passing and gives us coverage on all versions.

PR #247, opened 2020-07-29
Switch to OkHttp for networking

ZacSweers:
## Goal

Apache HTTP is largely deprecated. OkHttp is a reliable industry standard and kotlin-friendly API

Part of this involved starting to move the plugin extension to use more modern gradle property APIs for lazier configuration and a few opportunistic cleanups elsewhere

## Tests

Existing tests should continue working as regression tests

### DIFF
--- a/src/main/kotlin/com.bugsnag.android.gradle/BugsnagMultiPartUploadRequest.kt
+++ b/src/main/kotlin/com.bugsnag.android.gradle/BugsnagMultiPartUploadRequest.kt
@@ -155,6 +155,6 @@ internal interface BugsnagService {
     @POST
     fun uploadFile(
         @Url endpoint: String,
-        @PartMap parts: Map<String, RequestBody>
+        @PartMap parts: Map<String, @JvmSuppressWildcards RequestBody>
     ): Response<String>
 }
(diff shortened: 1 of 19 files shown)


ZacSweers:
I suspect this might fail tests as I haven't properly tested locally yet, but wanted to start the PR early and will continue iterating from here

src/main/kotlin/com.bugsnag.android.gradle/BugsnagMultiPartUploadRequest.kt
```diff
@@ -101,11 +109,60 @@ class BugsnagMultiPartUploadRequest {
...
+        private fun createService(timeoutDuration: Duration): BugsnagService {
+            return createService(OkHttpClient.Builder()
+                .connectTimeout(timeoutDuration)
+                .callTimeout(timeoutDuration)
+                .build())
+        }
+
+        internal fun createService(client: OkHttpClient): BugsnagService {
+            return Retrofit.Builder()
+                .baseUrl("https://example.com") // Not actually used
```
fractalwrench, line 141: would it be possible to default this to the value of `bugsnag.endpoint`, just on the off chance that a future refactor does alter this so that the baseUrl is used?

ZacSweers: Another, possibly more idiomatic way, would be to use this and then just make the path for the endpoint be "/"

phillipsam:

PR #249, opened 2020-08-03
Add autoUpdateBuildUuid flag to prevent manifest UUID generation

fractalwrench:
## Goal

Adds the `autoUpdateBuildUuid` flag to the `BugsnagPluginExtension` interface. This flag controls whether a UUID should be written to the AndroidManifest or not, as disabling this behaviour can improve performance by avoiding invalidation of Gradle's up-to-date checks. This flag remains `true` by default.

## Changeset

- Added flag to `BugsnagPluginExtension`
- Altered the build UUID to a placeholder string if `autoUpdateBuildUuid` is false, which results in the versionName/versionCode/applicationId being used to distinguish the mapping file instead, as per the [API docs](https://docs.bugsnag.com/api/android-mapping-upload/#for-proguard)

## Tests

- Added unit tests to verify behaviour when reading/writing a manifest with no uuid
- Added E2E scenario where `autoUpdateBuildUuid` is set to false
- Relies on the following tweak to mazerunner assertions: https://github.com/bugsnag/maze-runner/pull/112

### DIFF
--- a/src/main/kotlin/com.bugsnag.android.gradle/BugsnagPluginExtension.kt
+++ b/src/main/kotlin/com.bugsnag.android.gradle/BugsnagPluginExtension.kt
@@ -7,7 +7,6 @@ import org.gradle.api.provider.MapProperty
 import org.gradle.api.provider.Property
 import org.gradle.util.ConfigureUtil
 import java.io.File
-import javax.inject.Inject
 
 // To make kotlin happy with gradle's nullability
 private val NULL_STRING: String? = null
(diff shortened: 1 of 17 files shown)


src/main/kotlin/com.bugsnag.android.gradle/BugsnagPluginExtension.kt
```diff
@@ -25,6 +24,7 @@ open class BugsnagPluginExtension(objects: ObjectFactory) {
     var isUploadNdkMappings: Boolean? = null
     var isReportBuilds = true
     var isUploadDebugBuildMappings = false
+    var isAutoUpdateBuildUuid = true
```
ZacSweers, line 27: Any reason this shouldn't be `Property<Boolean>`? Then it could be zipped with the buildUuid provider property or otherwise piped in as a property to tasks

fractalwrench: Good point - this has been updated to `Property<Boolean>`.

PR #251, opened 2020-08-04
Convert plugin extension to use property syntax

fractalwrench:
## Goal

Converts the `bugsnag` plugin extension to use Gradle's [Property API](https://docs.gradle.org/current/javadoc/org/gradle/api/provider/Property.html). This conversion is required for future work to support [Configuration caching](https://docs.gradle.org/current/userguide/configuration_cache.html). The changeset is a breaking change depending on how users have configured bugsnag - upgrade steps are documented in the Upgrade guide.

## Changeset

- Made all public properties on `BugsnagPluginExtension` use the `Property` container object
- Reduced the visibility of task properties annotated with `@Internal` to allow non-breaking changes in future
- Converted task inputs to use properties where possible
- Ensured `PluginExtensionTest` unit test covers all default values

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -59,7 +59,7 @@ bugsnag {
 
 Previously the `bugsnag` plugin extension did not use Gradle's 
 [Property API](https://docs.gradle.org/current/javadoc/org/gradle/api/provider/Property.html),
-which meant the plugin could be configured like thus:
+which meant the plugin could be configured like this:
 
 ```groovy
 // old API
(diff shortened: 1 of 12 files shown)


UPGRADING.md
```diff
@@ -55,6 +55,29 @@ bugsnag {
 }
 ```
 
+#### `bugsnag` plugin extension uses Property API
+
+Previously the `bugsnag` plugin extension did not use Gradle's 
+[Property API](https://docs.gradle.org/current/javadoc/org/gradle/api/provider/Property.html),
+which meant the plugin could be configured like thus:
```
imjoehaines, line 62:
```suggestion
which meant the plugin could be configured like this:
```

PR #256, opened 2020-08-08
Use maven publish plugin for central publishing

### LINKED ISSUE #267: Missing sources in 5.0.0-alpha1
Looks like this is just due to the switch to Kotlin. #256 will help this for those consuming from maven since it configures Dokka for you, but I'm not sure the gradle plugin publisher tool speaks Dokka

image

ZacSweers:
This simplifies setup using the gradle maven publish plugin for publishing to central. This removes jcenter publishing, but publishing just to central covers jcenter since it proxies it anyway.

https://github.com/vanniktech/gradle-maven-publish-plugin

This requires added local gradle properties (recommended to put them in your ~/.gradle/gradle.properties) for signing and auth, described here: https://github.com/vanniktech/gradle-maven-publish-plugin#where-to-upload-to

Most significantly, this automatically handles publishing the gradle plugin marker artifact as well so that it can be used directly in `plugins {...}`

This also makes snapshots development (`./gradlew uploadArchives` when the version ends in `-SNAPSHOT`) and local development `./gradlew install`) easier

Resolves #267


ZacSweers:
This is... an interesting error

```
* What went wrong:
A problem occurred configuring root project 'app'.
> Could not resolve all artifacts for configuration ':classpath'.
   > Could not find com.bugsnag:bugsnag-android-gradle-plugin:9000.0.0-test.
     Searched in the following locations:
       - file:/home/travis/.m2/repository/com/bugsnag/bugsnag-android-gradle-plugin/9000.0.0-test/bugsnag-android-gradle-plugin-9000.0.0-test.pom
       - file:/home/travis/.m2/repository/com/bugsnag/bugsnag-android-gradle-plugin/9000.0.0-test/bugsnag-android-gradle-plugin-9000.0.0-test.jar
       - https://dl.google.com/dl/android/maven2/com/bugsnag/bugsnag-android-gradle-plugin/9000.0.0-test/bugsnag-android-gradle-plugin-9000.0.0-test.pom
       - https://dl.google.com/dl/android/maven2/com/bugsnag/bugsnag-android-gradle-plugin/9000.0.0-test/bugsnag-android-gradle-plugin-9000.0.0-test.jar
       - https://jcenter.bintray.com/com/bugsnag/bugsnag-android-gradle-plugin/9000.0.0-test/bugsnag-android-gradle-plugin-9000.0.0-test.pom
       - https://jcenter.bintray.com/com/bugsnag/bugsnag-android-gradle-plugin/9000.0.0-test/bugsnag-android-gradle-plugin-9000.0.0-test.jar
     Required by:
         project :
```

build.gradle
```diff
@@ -81,95 +79,23 @@ dependencies {
...
-            artifact jar
-            groupId "com.bugsnag"
-            artifactId "bugsnag-android-gradle-plugin"
-            version project.version
-        }
-    }
-}
-
-// Bintray upload
-bintray {
```
fractalwrench, line 156: One of our requirements is publishing the plugin to mavenCentral, plugins.gradle, and bintray. Is there an equivalent way of publishing to bintray when using the maven publish plugin?

ZacSweers: JCenter will source maven central to mirror anything it doesn't have. I think the plug-in could handle bintray uploads but would need to check. Does the mirroring not cover the bintray need?

fractalwrench: We've discussed this internally and are happy to drop publishing to bintray. The only reason it was added was that several years ago we had issues with mavenCentral's reliability, which haven't been experienced recently. We'll probably hold off merging this until the next major of the bugsnag gradle plugin is released to reduce the scope of changes - hopefully this should be quite soon.

ZacSweers:
Looks like it's working now 👍 . Would be worth exploring enabling snapshots btw, makes it easy for people to try new versions. This plugin handles it well too as long at the version ends in `-SNAPSHOT`

Issue #201, opened 2020-04-30
Please create Bugsnag tasks using Gradle's creation avoidance API


rallat:
### Expected behavior
Create bugsnag tasks lazily using this API
https://docs.gradle.org/current/userguide/task_configuration_avoidance.html
Rather than [creating them](https://github.com/bugsnag/bugsnag-android-gradle-plugin/blob/master/src/main/groovy/com/bugsnag/android/gradle/BugsnagPlugin.groovy#L125) you can [register them](https://docs.gradle.org/current/userguide/task_configuration_avoidance.html#sec:task_configuration_avoidance_general) 

### Observed behavior
Plugin should lazily create tasks to avoid hitting configuration time.
Our build scan shows that the bugsnag plugin is creating 921 task at configuration time. However, most of them are for variants not related to the task running so I believe they could be registered instead of created as shared above.
```
com.bugsnag.android.gradle | Plugin | 0.186s |   | Applied to 1 project | 916 tasks created
```

### Steps to reproduce
Run a build `--scan` https://guides.gradle.org/creating-build-scans/
and go to Performance > configuration > search for `com.bugsnag.android.gradle`
For us it creates 921 tasks the creation can be deferred to when they are actually needed

### Version
4.22.3

### Additional information
*[Insert any additional information]*

mattdyoung:
Hi @rallat 

Thanks for the suggestion! This sounds like a useful improvement to our gradle plugin so we'll take a look at doing this when we have the bandwidth.

ZacSweers:
This is mostly done on the `v5` branch, there's just one little problem area in the plugin's `registerBugsnagTasksForVariant()` logic that needs cleaning up. I've been talking with @fractalwrench a bit about it, started down that path but haven't quite gotten it working just yet. Here's a patch of what I got to so far

```patch
Index: src/main/kotlin/com.bugsnag.android.gradle/MappingFileProvider.kt
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- src/main/kotlin/com.bugsnag.android.gradle/MappingFileProvider.kt	(revision 714a6e91f9fbdefefe9494fb3b3d19e492e95a8e)
+++ src/main/kotlin/com.bugsnag.android.gradle/MappingFileProvider.kt	(date 1596949593000)
@@ -3,9 +3,7 @@
 import com.android.build.gradle.AppExtension
[...]

PR #257, opened 2020-08-09
Support configuration caching

### LINKED ISSUE #253: Support Gradle Configuration Caching
@fractalwrench and I talked about this offline, filing to track here!

https://docs.gradle.org/nightly/userguide/configuration_cache.html

ZacSweers:
Resolves #253

~There is still more work to do though, namely `UploadRequestClient` needs to be rethought as its state-sharing nature is incompatible with configuration caching. This PR gets most of the way there though!~

Fully implements support for configuration caching, and modernizes much of the internals along the way in a backward-compatible way!

Some highlights:
* Configuration caching support
* Uses a `BuildService` to share state for certain things, like `UploadRequestClient` and okhttpclient instances
* Optimized I/O bits in certain places using Okio
* Backported/backward-compatible patterns for tasks to use new Gradle features as needed


ZacSweers:
The test failure appears to be due to the tweaked hashcode calculation to exclude buildUuid like the doc mentioned. Reverted that for now in dcf8007 but probably worth a second look later

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,5 @@
-#Mon Jul 06 17:13:48 BST 2020
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.5.1-all.zip
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-zipStorePath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.6-rc-6-all.zip
```
ZacSweers, line 3: This is for access to `ArchiveOperations`

src/main/kotlin/com.bugsnag.android.gradle/AndroidManifestInfo.kt
```diff
@@ -9,11 +9,11 @@ import java.io.File
...
-    var apiKey: String,
-    var versionCode: String,
-    var buildUUID: String?,
-    var versionName: String,
-    var applicationId: String
+    val apiKey: String,
+    val versionCode: String,
+    val buildUUID: String?,
+    val versionName: String,
+    val applicationId: String
```
ZacSweers, line 16: I assumed these were unintentionally `var`

src/main/kotlin/com.bugsnag.android.gradle/AndroidManifestParser.kt
```diff
@@ -63,6 +62,8 @@ class AndroidManifestParser {
                     |versionCode=$versionCode
                     |versionName=$versionName
                     |applicationId=$applicationId
+                    |Manifest file = $manifestPath
+                    |Manifest content=${manifestPath.readText()}
```
ZacSweers, line 66: I accidentally left this in from debugging, but kind of think at least printing the file is useful for debugging in general too. Will remove the content print, let me know if you want to remove the file path too

```suggestion
```

src/main/kotlin/com.bugsnag.android.gradle/AndroidManifestParser.kt
```diff
@@ -13,11 +13,10 @@ import java.io.IOException
 import java.io.PrintWriter
 import javax.xml.parsers.ParserConfigurationException
 
-class AndroidManifestParser {
+internal class AndroidManifestParser {
 
     private val namespace = Namespace("http://schemas.android.com/apk/res/android", "android")
 
-    @Throws(ParserConfigurationException::class, SAXException::class, IOException::class)
```
ZacSweers, line 20: Assumed this was leftover from java -> kotlin conversion

src/main/kotlin/com.bugsnag.android.gradle/BugsnagFileUploadTask.kt
```diff
@@ -8,6 +9,7 @@ interface BugsnagFileUploadTask {
     val endpoint: Property<String>
     val retryCount: Property<Int>
     val timeoutMillis: Property<Long>
+    val httpClientHelper: Property<BugsnagHttpClientHelper>
```
ZacSweers, line 12: This is a nice benefit of using the interface - we can provide this dynamically depending on gradle version in the plugin and tasks know nothing about its implementation

src/main/kotlin/com.bugsnag.android.gradle/BugsnagInstallJniLibsTask.kt
```diff
@@ -60,5 +74,93 @@ open class BugsnagInstallJniLibsTask @Inject constructor(
...
+                  project.tasks.register<BugsnagInstallJniLibsTask53Plus>(name, configurationAction)
+              }
+              else -> {
+                  project.tasks.register<BugsnagInstallJniLibsTaskLegacy>(name, configurationAction)
+              }
+            }
+        }
     }
 }
+
```
ZacSweers, line 109: I think the docs and naming below are mostly self explanatory, but one thing I thought was worth mentioning is that it might be better to use inheritance to de-dupe some of the logic here (and in other task facades). Let me know what you think, I'm sort of undecided since the duplication isn't that much

src/main/kotlin/com.bugsnag.android.gradle/BugsnagInstallJniLibsTask.kt
```diff
@@ -60,5 +74,93 @@ open class BugsnagInstallJniLibsTask @Inject constructor(
...
+/**
+ * A Gradle 6.6+ compatible [BugsnagInstallJniLibsTask], which uses [FileSystemOperations]
+ * and [ArchiveOperations] to support configuration caching.
+ */
+internal open class BugsnagInstallJniLibsTaskGradle66Plus @Inject constructor(
+    objects: ObjectFactory,
+    projectLayout: ProjectLayout,
+    private val fsOperations: FileSystemOperations,
+    private val archiveOperations: ArchiveOperations
```
ZacSweers, line 159: This is what necessitated Gradle 6.6. May be worth adding a 6.6 matrix entry in tests

src/main/kotlin/com.bugsnag.android.gradle/BugsnagManifestUuidTask.kt
```diff
@@ -59,10 +60,10 @@ open class BugsnagManifestUuidTask @Inject constructor(objects: ObjectFactory) :
     fun updateManifest() {
         val manifestPath = getManifestPaths(project, variant, variantOutput)
         if (manifestPath == null) {
-            project.logger.warn("Bugsnag: Failed to find manifest at $manifestPath for $variantOutput")
+            logger.warn("Bugsnag: Failed to find manifest at $manifestPath for $variantOutput")
```
ZacSweers, line 63: While this task isn't compatible with configuration caching (not an issue since it wouldn't be used on versions that support it anyway), I did clean this up just to make it more consistent with other tasks

src/main/kotlin/com.bugsnag.android.gradle/BugsnagMultiPartUploadRequest.kt
```diff
@@ -37,10 +36,10 @@ class BugsnagMultiPartUploadRequest(
     private val overwrite: Boolean,
     private val endpoint: String,
     private val retryCount: Int,
-    timeoutDuration: Duration
+    okHttpClient: OkHttpClient
 ) {
 
-    private val bugsnagService = createService(timeoutDuration)
+    private val bugsnagService = createService(okHttpClient)
```
ZacSweers, line 42: Was very happy with how nicely this played with our previously set up overloads 😁

src/main/kotlin/com.bugsnag.android.gradle/BugsnagPlugin.kt
```diff
@@ -57,6 +61,24 @@ class BugsnagPlugin : Plugin<Project> {
...
+            val httpClientHelperProvider = if (canUseBuildService) {
+                project.gradle.sharedServices.registerIfAbsent("bugsnagHttpClientHelper",
+                    BuildServiceBugsnagHttpClientHelper::class.java
+                ) { spec ->
+                    // Provide some parameters
+                    spec.parameters.timeoutMillis.set(bugsnag.requestTimeoutMs)
+                }
+            } else {
+                // Reuse instance
+                val client = LegacyBugsnagHttpClientHelper(bugsnag.requestTimeoutMs)
```
ZacSweers, line 74: Note the `OkHttpClient` itself is lazily initialized internally in this class

src/main/kotlin/com.bugsnag.android.gradle/BugsnagPlugin.kt
```diff
@@ -83,28 +105,40 @@ class BugsnagPlugin : Plugin<Project> {
...
+        project: Project,
+        bugsnag: BugsnagPluginExtension,
+        android: AppExtension
+    ) {
         val ndkTasks = project.tasks.withType(ExternalNativeBuildTask::class.java)
         val cleanTasks = ndkTasks.filter { it.name.contains(CLEAN_TASK) }.toSet()
         val buildTasks = ndkTasks.filter { !it.name.contains(CLEAN_TASK) }.toSet()
 
         if (buildTasks.isNotEmpty()) {
-            val ndkSetupTask = project.tasks.create("bugsnagInstallJniLibsTask", BugsnagInstallJniLibsTask::class.java)
```
ZacSweers, line 101: This helps with #201

src/main/kotlin/com.bugsnag.android.gradle/BugsnagPlugin.kt
```diff
@@ -83,28 +105,40 @@ class BugsnagPlugin : Plugin<Project> {
...
-            ndkSetupTask.buildDirDestination.set(File(project.buildDir, "/intermediates/bugsnag-libs"))
-            val files = resolveBugsnagArtefacts(project)
-            ndkSetupTask.bugsnagArtefacts.set(files)
+            val ndkSetupTask = BugsnagInstallJniLibsTask.register(project, "bugsnagInstallJniLibsTask") {
+                val files = resolveBugsnagArtifacts(project)
+                bugsnagArtifacts.from(files)
```
ZacSweers, line 135: Just something I wanted to mention - `ConfigurableFileCollection` should conventionally use `from()` rather than `Property.set()`. Fixed these up during this

src/main/kotlin/com/bugsnag/android/gradle/internal/GradleUtil.kt
```diff
@@ -0,0 +1,92 @@
...
+    val VERSION_5_3: VersionNumber = VersionNumber.parse("5.3")
+    val VERSION_6: VersionNumber = VersionNumber.parse("6.0")
+    val VERSION_6_1: VersionNumber = VersionNumber.parse("6.1")
+    val VERSION_6_6: VersionNumber = VersionNumber.parse("6.6")
+}
+
+internal fun Gradle.versionNumber(): VersionNumber = VersionNumber.parse(gradleVersion)
+
+/** A fast file hash that don't load the entire file contents into memory at once. */
+internal fun File.md5HashCode(): Int {
```
ZacSweers, line 30: This should be a significant memory win as it never loads the entire file into memory. Some of our mapping files are 70MB+ (R8 generates huge files), so loading this all into memory was 🥵 .

src/main/kotlin/com/bugsnag/android/gradle/internal/GradleUtil.kt
```diff
@@ -0,0 +1,92 @@
...
+    return HashingSource.md5(source()).use { hashingSource ->
+        hashingSource.buffer().readAll(blackholeSink())
+        hashingSource.hash.hashCode()
+    }
+}
+
+/* Borrowed helper functions from the Gradle Kotlin DSL. */
```
ZacSweers, line 37: I believe you can just use kotlin-dsl directly without imposing it on your downstream consumers, but this subset of APIs was easy enough to just copy in for now

src/main/kotlin/com/bugsnag/android/gradle/internal/GradleUtil.kt
```diff
@@ -0,0 +1,92 @@
+@file:Suppress("MatchingDeclarationName") // This file contains multiple top-level members
+package com.bugsnag.android.gradle.internal
```
ZacSweers, line 2: I mentioned this to @fractalwrench separately, but I noticed that the plugin's regular package is actually one file with dots in the name

image

src/main/kotlin/com/bugsnag/android/gradle/internal/UploadRequestClient.kt
```diff
@@ -0,0 +1,67 @@
...
+
+    /**
+     * Executes request with automatic de-duplication of similar requests.
+     *
+     * If the version information and payload match another request that
+     * has already been enqueued, the response value of the existing request
+     * will be returned. If no existing requests match, a new request will
+     * be enqueued and executed.
+     *
+     * Equality is measured by the [AndroidManifestInfo] (excluding buildUuid)
```
ZacSweers, line 24: This doesn't actually exclude the buildUuid property. I tried to exclude it in f1c443b8b2793375f988632edb01437acbe0639d but that resulted in some failing tests that expected more results than would happen when it's excluded. Reverted and left for another time

src/main/kotlin/com.bugsnag.android.gradle/AndroidManifestInfo.kt
fractalwrench: Yep - I don't think there's any reason for these to not be `val`.

src/main/kotlin/com.bugsnag.android.gradle/AndroidManifestParser.kt
fractalwrench: Yep, this is the case.

src/main/kotlin/com/bugsnag/android/gradle/internal/GradleUtil.kt
fractalwrench: I've raised an internal ticket for this and will address this once the open PRs have been merged to avoid any potential for merge conflicts.

src/main/kotlin/com/bugsnag/android/gradle/internal/UploadRequestClient.kt
fractalwrench: The docs are wrong on this - the buildUuid property should be part of the hashcode. There's been a bit of back and forth on the design behind this internally, but we've now settled on the following behaviour:

- By default the plugin generates a build UUID, and it is recommended that the plugin should be disabled for any unnecessary variants to improve build performance
- Advanced users can specify their own deterministic build UUID by adding a `meta-data` tag in the manifest

It's a bit of a tricky balance between the risk of overwriting existing mapping files with the same version information, versus improving performance by using cached task outputs. This approach is the best solution we've come up with so far.

In terms of actions, the comment just needs editing here.

fractalwrench, line 24:
```suggestion
     * Equality is measured by the [AndroidManifestInfo]
```

src/main/kotlin/com/bugsnag/android/gradle/internal/UploadRequestClient.kt
```diff
@@ -0,0 +1,67 @@
...
+import com.bugsnag.android.gradle.AndroidManifestInfo
+import org.gradle.api.Project
+import org.gradle.api.provider.Provider
+import org.gradle.api.services.BuildService
+import org.gradle.api.services.BuildServiceParameters
+import java.util.Objects
+import java.util.concurrent.ConcurrentHashMap
+import java.util.concurrent.FutureTask
+
+sealed class UploadRequestClient : AutoCloseable {
```
fractalwrench, line 12: nice use of a sealed class 👍

src/main/kotlin/com/bugsnag/android/gradle/internal/GradleUtil.kt
fractalwrench: Looks good :) if extra functionality is required in future it'd be worth setting up the kotlin-dsl

src/main/kotlin/com/bugsnag/android/gradle/internal/GradleUtil.kt
```diff
@@ -0,0 +1,92 @@
...
+import org.gradle.api.provider.SetProperty
+import org.gradle.api.tasks.TaskContainer
+import org.gradle.api.tasks.TaskProvider
+import org.gradle.util.VersionNumber
+import java.io.File
+
+internal object GradleVersions {
+    val VERSION_5_3: VersionNumber = VersionNumber.parse("5.3")
+    val VERSION_6: VersionNumber = VersionNumber.parse("6.0")
+    val VERSION_6_1: VersionNumber = VersionNumber.parse("6.1")
```
fractalwrench, line 23: nit: `SYS_PROPERTIES_VERSION` in `BugsnagReleasesTask` could use this constant instead

src/main/kotlin/com.bugsnag.android.gradle/BugsnagInstallJniLibsTask.kt
fractalwrench: I've created an internal ticket to track this - running against 6.6 and AGP 4.2.0 canary seems like a sensible idea.

src/main/kotlin/com.bugsnag.android.gradle/BugsnagInstallJniLibsTask.kt
```diff
@@ -60,5 +74,93 @@ open class BugsnagInstallJniLibsTask @Inject constructor(
...
+        internal fun register(
+            project: Project,
+            name: String,
+            configurationAction: BugsnagInstallJniLibsTask.() -> Unit
+        ): TaskProvider<out BugsnagInstallJniLibsTask> {
+            val gradleVersion = project.gradle.versionNumber()
+            return when {
+              gradleVersion >= GradleVersions.VERSION_6 -> {
+                  when {
+                    gradleVersion >= GradleVersions.VERSION_6_6 -> {
```
fractalwrench, line 91: style nit: the indentation is off by two spaces for this `when` block

src/main/kotlin/com.bugsnag.android.gradle/BugsnagInstallJniLibsTask.kt
fractalwrench: I think the current approach works well and it's quite clear that each task supports a range of Gradle versions. A bit of duplication is fine particularly as there are only a couple of subclasses at the moment.

PR #261, opened 2020-08-11
Avoid eager task initialization

### LINKED ISSUE #201: Please create Bugsnag tasks using Gradle's creation avoidance API

ZacSweers:
Resolves #201 by avoiding eagerly initializing other tasks.

Still needs more work but sticking this up for early feedback. Will continue iterating on this! Commits get a little messy due to things being interconnected :|


src/main/kotlin/com/bugsnag/android/gradle/BugsnagPlugin.kt
```diff
@@ -170,48 +170,44 @@ class BugsnagPlugin : Plugin<Project> {
...
-            symbolFileTask?.get()?.let { task ->
-                val ndkSearchDirs = symbolFileTask.get().searchDirectories
-                releasesTask.get().ndkMappingFileProperty.from(ndkSearchDirs)
-                task.uploadRequestClient.set(ndkUploadClientProvider)
-            }
-            releasesTask.get().uploadRequestClient.set(releasesUploadClientProvider)
-            releasesTask.get().manifestInfoFile.set(manifestInfoFile)
-            symbolFileTask?.get()?.manifestInfoFile?.set(manifestInfoFile)
-            releasesTask.get().manifestInfoFile.set(manifestInfoFile)
+            }?.registerWithAssembleTasks(variant, output, bugsnag, project.provider { true })
```
fractalwrench, line 199: I think this should probably be `project.provider { bugsnag.uploadNdkMappings }`

src/main/kotlin/com/bugsnag/android/gradle/BugsnagPlugin.kt
```diff
@@ -283,11 +284,12 @@ class BugsnagPlugin : Plugin<Project> {
         return BugsnagUploadNdkTask.register(project, taskName) {
             this.requestOutputFile.set(requestOutputFile)
             projectRoot.set(bugsnag.projectRoot.getOrElse(project.projectDir.toString()))
-            searchDirectories.from(getSearchDirectories(project, variant))
+            searchDirectories.from()
```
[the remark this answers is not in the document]
ZacSweers: accident 😬

src/main/kotlin/com/bugsnag/android/gradle/BugsnagPlugin.kt
ZacSweers: Good call. It was hardcoded to `true` before so I was just matching this originally

phillipsam:

PR #266, opened 2020-08-14
Avoid eager initialization of bugsnag tasks

fractalwrench:
## Goal

Avoids eager initialization of bugsnag tasks by supporting [Task Configuration Avoidance](https://docs.gradle.org/current/userguide/task_configuration_avoidance.html). Previous changesets such as #227 made pre-requisite changes, but the use of `task.get()` triggered creation of bugsnag tasks during the Configuration phase which impacts build performance.

## Changeset

This changeset builds on #261 which has been reviewed by @fractalwrench. The majority of the new changes in this PR relate to the E2E scenarios as the request order has changed due to the different mechanism used to register bugsnag tasks.

`BugsnagUploadProguardTask` has been updated to use a `ConfigurableFileCollection` and `@InputFiles` for the mapping file property. These do not require a file to exist at the time of task registration - which would otherwise fail the build as the mapping file doesn't exist until late in the build lifecycle. This mirrors the approach already used for `BugsnagReleasesTask`.

On AGP 3.5.X the tasks add a `mustRunAfter` dependency on the package task, as the ordering is otherwise incorrect and the mapping file would not exist. This leads to a degradation as configuration avoidance can't be supported, and also results in manual invocations of the upload/bundle upload tasks failing. This is seen as acceptable due to the low number of users on this version, the less common use-case, and a simple workaround of updating AGP version.

## Tests

- Updated E2E scenarios to deal with new request order
- Skipped manual upload scenarios on AGP 3.5.0 for graceful degradation of behaviour
- Ran a gradle [build scan](https://guides.gradle.org/creating-build-scans/) and confirmed that only 2 tasks are created by the bugsnag plugin during the configuration phase, whereas previously 220 were created


fractalwrench:
PLAT-4842 should cover allowing the tests to be enabled once more - this is just an issue with the order in which requests are received on particular AGP versions.

PR #269, opened 2020-08-18
Respect bugsnag.retryCount by using OkHttp interceptor

fractalwrench:
## Goal

#247 switches the internal HTTP library to OkHttp to avoid using the deprecated Apache client. This caused a bug where the `bugsnag.retryCount` property was not respected, and if `failOnUploadError` was set then the request always failed after the first attempt.

This changeset fixes the issue by using a more idiomatic approach of retrying requests with an [OkHttp Interceptor](https://square.github.io/okhttp/interceptors/). This allows for the legacy `Call` class to be removed. Exceptions thrown by the request are also now caught, and only rethrown if the `failOnUploadError` flag is set to true.

Tested by running each task pointed at a local netcat server, and verifying that the request was retried the number of times set on `retryCount`. If `failOnUploadError` was set to true then the build failed with an abnormal exit code; if it was set to false then the build succeeded.


src/main/kotlin/com/bugsnag/android/gradle/BugsnagPluginExtension.kt
```diff
@@ -50,7 +50,7 @@ open class BugsnagPluginExtension(objects: ObjectFactory) {
         .convention(false)
 
     val retryCount: Property<Int> = objects.property<Int>()
-        .convention(0)
+        .convention(1)
```
imjoehaines, line 53: I think this should stay as 0 because 0 should mean "attempt the request once, but don't try again if it fails". From reading the code it seems like `retryCount` has become `maxRequestCount`, which is a little bit different

src/main/kotlin/com/bugsnag/android/gradle/internal/BugsnagHttpClientHelper.kt
```diff
@@ -25,23 +27,55 @@ abstract class BuildServiceBugsnagHttpClientHelper
...
+            while (response == null && maxRetries >= attempts) {
+                try {
+                    val request = chain.request()
+                    response = chain.proceed(request)
+                } catch (exc: Throwable) {
+                    cause = exc
+                }
+                attempts++
+            }
+            return response ?: throw IllegalStateException("Bugsnag request failed to complete", cause)
```
imjoehaines, line 77: I think this would be a little clearer with a do/while loop, because we always want the request to happen at least once (which is currently guaranteed with `response == null`) and then a `maxRetries` of 0 works as expected

```suggestion
            var attempts = 0
            var cause: Throwable? = null

            do {
                try {
                    val request = chain.request()

                    return chain.proceed(request)
                } catch (exc: Throwable) {
                    cause = exc
                }
                
                attempts++
            } while (attempts < maxRetries)

            throw IllegalStateException("Bugsnag request failed to complete", cause)
```

fractalwrench: This is clearer - I've adopted this suggestion.

PR #200, opened 2020-04-25
Support uploading via proxy connections if proxy system properties are set

jrodbx:
Closes #199

## Goal
Details in #199 

## Design
Allows opt-in support for proxy connections

## Changeset
The uploading client setup in `BugsnagMultiPartUploadTask`

## Tests
No tests currently exist for this task.  Instead, I first tested locally by installing to mavenLocal and debugged the plugin's execution using `-Dorg.gradle.debug=true` and remote attaching a debugger.  Then, I pushed to our internal artifact repository and tested a CI build.  A different error message (403 forbidden) resulted since we currently don't have a proxy whitelist entry for  `upload.bugsnag.com`.  However, the proxy connection and authentication issues have been resolved.

PR #270, opened 2020-08-18
Support proxies via OkHttp

ZacSweers:
Continuation of #200 but implemented in the new OkHttp stack. This also opportunistically fixes the `BugsnagReleasesTask` not using the same shared client

### DIFF
--- a/src/main/kotlin/com/bugsnag/android/gradle/internal/BugsnagHttpClientHelper.kt
+++ b/src/main/kotlin/com/bugsnag/android/gradle/internal/BugsnagHttpClientHelper.kt
@@ -42,10 +42,12 @@ interface BugsnagHttpClientHelper : AutoCloseable {
             project: Project,
             bugsnag: BugsnagPluginExtension
         ): Provider<out BugsnagHttpClientHelper> {
-            val proxyHost = project.providers.systemPropertyCompat(KEY_PROXY_HOST)
-            val proxyPort = project.providers.systemPropertyCompat(KEY_PROXY_PORT)
-            val proxyUsername = project.providers.systemPropertyCompat(KEY_PROXY_USER)
-            val proxyPassword = project.providers.systemPropertyCompat(KEY_PROXY_PASSWORD)
+            val gradleVersion = project.gradle.versionNumber()
+            val providers = project.providers
+            val proxyHost = providers.systemPropertyCompat(KEY_PROXY_HOST, gradleVersion)
+            val proxyPort = providers.systemPropertyCompat(KEY_PROXY_PORT, gradleVersion)
+            val proxyUsername = providers.systemPropertyCompat(KEY_PROXY_USER, gradleVersion)
+            val proxyPassword = providers.systemPropertyCompat(KEY_PROXY_PASSWORD, gradleVersion)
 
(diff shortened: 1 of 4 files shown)


src/main/kotlin/com/bugsnag/android/gradle/internal/BugsnagHttpClientHelper.kt
```diff
@@ -19,33 +36,102 @@ interface BugsnagHttpClientHelper : AutoCloseable {
...
+        fun create(
+            project: Project,
+            bugsnag: BugsnagPluginExtension
+        ): Provider<out BugsnagHttpClientHelper> {
+            val gradleVersion = project.gradle.versionNumber()
+            val providers = project.providers
+            val proxyHost = providers.systemPropertyCompat(KEY_PROXY_HOST, gradleVersion)
+            val proxyPort = providers.systemPropertyCompat(KEY_PROXY_PORT, gradleVersion)
+            val proxyUsername = providers.systemPropertyCompat(KEY_PROXY_USER, gradleVersion)
+            val proxyPassword = providers.systemPropertyCompat(KEY_PROXY_PASSWORD, gradleVersion)
```
swankjesse, line 50: If you didn’t need system properties, you could have probably just used the JVM’s built-in stuff that uses these properties for you.

```
    OkHttpClient.Builder()
        .proxySelector(ProxySelector.getDefault())
        .proxyAuthenticator(okhttp3.Authenticator.JAVA_NET_AUTHENTICATOR)
        ...
        .build()
```

ZacSweers: That's a good point! Gradle reaaaaally wants to have all these property lookups run through them for configuration caching and build hermeticity though. Will go with this for now and I'll raise this up with gradle people separately to see how much it matters when they're not being used as task inputs. Easy to change later since this PR makes all these internal APIs now

phillipsam:
Thanks for the PR @ZacSweers! We'll take a look

src/main/kotlin/com/bugsnag/android/gradle/internal/BugsnagHttpClientHelper.kt
ZacSweers: Checked in the gradle slack, looks like it's ok to use this since it doesn't affect configuration of tasks. Will update 👍

ZacSweers: Updated here 63bb67e. There may still be some potential use if, say, a user wanted to specify these via the plugin extension. I think that seems worth waiting until someone actually asks for it though

ZacSweers:
the CI failure seems like it might be a flake? Let me merge v5 to re-kick

fractalwrench:
Hey @ZacSweers - we're happy to accept this and are currently planning out how to test this in an automated way with the mazerunner framework. If you could target `next` we'll take a look at this once we've finished that design work.

mattdyoung:
Ended up being covered by https://github.com/bugsnag/bugsnag-android-gradle-plugin/pull/298 released in [v5.2.0](https://github.com/bugsnag/bugsnag-android-gradle-plugin/releases/tag/v5.2.0)

PR #276, opened 2020-08-21
Use maven publish plugin for central publishing

### LINKED ISSUE #267: Missing sources in 5.0.0-alpha1

ZacSweers:
Re-publish of #256

This simplifies setup using the gradle maven publish plugin for publishing to central. This removes jcenter publishing, but publishing just to central covers jcenter since it proxies it anyway.

https://github.com/vanniktech/gradle-maven-publish-plugin

This requires added local gradle properties (recommended to put them in your ~/.gradle/gradle.properties) for signing and auth, described here: https://github.com/vanniktech/gradle-maven-publish-plugin#where-to-upload-to

Most significantly, this automatically handles publishing the gradle plugin marker artifact as well so that it can be used directly in `plugins {...}`

This also makes snapshots development (`./gradlew uploadArchives` when the version ends in `-SNAPSHOT`) and local development `./gradlew install`) easier

Resolves #267


fractalwrench:
Hi Zac, I tried this out after resolving the conflicts against the `next` branch. Installing the plugin locally does seem to work and I can see javadoc is attached to the jars, but I'm running into trouble when running the `uploadArchives` or `bintrayUpload` task, namely:

>./gradlew uploadArchives                  `
Configuration(s) specified but the install task does not exist in project :.
Publication Publication not found in project :.

This seems to result in only the `bugsnag-android-gradle-plugin-5.X.X.jar` being uploaded to Sonatype, and similar behaviour for Bintray. I'm assuming that there's some difference in the vanniktech plugin that it doesn't pick up the rest of these files by default. I think this would be the only obstacle to merging this PR.

ZacSweers:
Let me take a look and test this out locally

ZacSweers:
@fractalwrench can you test again? I think the install task issue was due to a previous configuration being leftover that I removed in f6e7a85.

I see all the requisite files when I install locally 

image


I don't really understand these errors from the bintray plugin

```
Configuration(s) specified but the install task does not exist in project :.
Publication BintrayPublication not found in project :.
```

The bintray plugin is... not good. We use jfrog internally in combination with this plugin for publishing, could you try the latest change with `./gradlew uploadArchivesBintray`? Specify the user and key as you did before

fractalwrench:
This seems to work nicely when installing locally, after setting the sourceCompatibility to JDK 8 explicitly. It also appears to upload the correct artefacts to Sonatype although I've not tried resolving the dependency from the staging repository as of yet. The bintray publish task seems to have advanced onto a new error message:

>Execution failed for task ':publishPluginMavenPublicationToBintrayRepository'.
> Failed to publish publication 'pluginMaven' to repository 'bintray'
   > Could not PUT 'https://api.bintray.com/maven/fractalwrench/artifact-sandbox/;publish=1/com/bugsnag/bugsnag-android-gradle-plugin/5.3.0/bugsnag-android-gradle-plugin-5.3.0.jar'. Received status code 404 from server: Not Found

I tried altering the URL to use `bugsnag` rather than `$bintray_user` which advanced things a little further, but I got caught up on a similar error message:

> Execution failed for task ':publishPluginMavenPublicationToBintrayRepository'.
> Failed to publish publication 'pluginMaven' to repository 'bintray'
   > Could not PUT 'https://api.bintray.com/maven/bugsnag/artifact-sandbox/;publish=1/com/bugsnag/bugsnag-android-gradle-plugin/5.3.0/bugsnag-android-gradle-plugin-5.3.0.jar'. Received status code 404 from server: Not Found

`bintray_user` and `bintray_api_key` are both present in `~/.gradle/gradle.properties`, so I'm not too sure what's going on here.

Also I noticed that `publish=1` - I've altered this to `publish=0` as my understanding is that the first would publish the package automatically.

johnkiely1:
Hey @ZacSweers, 

Just to let you know this has now been released in [V5.7.7](https://github.com/bugsnag/bugsnag-android-gradle-plugin/releases/tag/v5.7.7). 
Thanks again for the PR.

Issue #274, opened 2020-08-21
Same mapping file is re-uploaded multiple times


ZacSweers:
### Expected behavior

I'd expect this to only upload once as it's identical in all of these ABI variants

### Observed behavior

It uploads the same file every time 
![image](https://user-images.githubusercontent.com/1361086/90850344-6ed32e80-e33f-11ea-9464-0864681e8d6d.png)

### Steps to reproduce
https://github.com/ZacSweers/CatchUp/pull/246

Run `./gradlew :app:assembleRelease`

### Version
5.0.0

### Additional information

Right now, this is intended behavior. It's quite wasteful though, and I think could be consolidated since these files are the same for every ABI. This is currently configured in BugsnagPlugin here: https://github.com/bugsnag/bugsnag-android-gradle-plugin/blob/36600757dae7a4a1979664ce67bb00fba5b084eb/src/main/kotlin/com/bugsnag/android/gradle/BugsnagPlugin.kt#L286

fractalwrench:
My understanding is that this was addressed in #248. Evidently the logged message is a bit misleading and should log something about the upload having been skipped instead, as [this message](https://github.com/bugsnag/bugsnag-android-gradle-plugin/blob/75a803ef393f1e5372a4d35c7c7c3d6507d5acbe/src/main/kotlin/com/bugsnag/android/gradle/BugsnagUploadProguardTask.kt#L105) will log whenever an upload occurs.

fractalwrench:
I think the proposed changes in #277 wouldn't work in the scenario where automatic upload is disabled, and a user invokes an individual upload task like this:  `./gradlew assemble uploadBugsnag${variantOutputName}Mapping`

As that's something we want to support and there doesn't seem to be another obvious way around this, I'll proceed with altering the log messages to make it a bit more evident that the upload is being skipped for most of these tasks.

phillipsam:
A fix for this has now been released in [v5.0.2](https://github.com/bugsnag/bugsnag-android-gradle-plugin/releases/tag/v5.0.2).

PR #277, opened 2020-08-21
 Only upload mapping files once per variant

ZacSweers:
## Goal

This is a proposal solution for #274, and works by only registering the mapping file upload at most one time. This is useful as these mapping files are identical per variant, and can save quite a bit of time in uploads for apps with large mapping files and/or multiple ABIs/densities

## Design

See above

## Changeset

See title

## Tests

I can look at updating the maze tests if y'all are ok with this proposal


ZacSweers:
`UploadRequestClient.makeRequestIfNeeded` is sort of supposed to handle this, but I think this is likely a better solution

phillipsam:
@ZacSweers thanks for the PR! We'll take a look

PR #281, opened 2020-08-22
Show a progress bar during mapping file uploads

ZacSweers:
## Goal

Mapping files often take a long time to upload, and while debugging I found myself wishing I could see the progress. This does that!

Screen Shot 2020-08-22 at 3 50 59 PM

## Design

This is something we do internally for downloads and was easily adapted for uploads by wrapping the request body if the retrofit endpoint opted it in via `@DisplayProgress`.

This can be disabled via `bugsnag { displayUploadProgress = false }` in gradle

This would go nicely with #277 as well. Right now it's only set up for mapping file uploads but can easily be set up for other endpoints by simply annotating them with that same `@DisplayProgress` annotation.

## Changeset

See above

## Tests

Added a simple smoke test and also manually tested in our own project


steve-nester-uk:
@ZacSweers thanks for raising the PR.  It's in the queue for a review, we'll revert here with any review comments in due course.

fractalwrench:
I've resolved the conflicts and tried this out locally and had a couple of queries. 

1. when running `mazerunner features/ndk_app_agp400.feature` locally, the upload message doesn't appear to display for the `BugsnagUploadProguardTask`, and only displays once rather than 4 times for the `BugsnagUploadNdkTask`.
2. Most of the time the console seems to get stuck on this last message which could be a bit misleading - is there any way to clear/reset that to indicate that the upload was completed? 
>Uploading  99% [===================================>] 0/0MB (0:00:00 / 0:00:00)

ZacSweers:
Sorry for the delay! 

For 1 - if the upload happens fast enough then it will sometimes not show at all. Could that maybe be what you were seeing?

For 2 - it implements `Closeable`, so if we can bind it to something like task execution we can make sure it gets closed. I can take a look at if there's an edge case where we don't close

PR #284, opened 2020-08-24
Make BuildUuid a settable extension property instead

ZacSweers:
## Goal

This allows consumers that do have their own deterministic heuristic to be able to set their own in a way that aligns with their own reproducibility/compliance requirements, as well as being able to better track down build UUIDs associated with previous releases.

## Design

This consolidates build uuid API usage in the manifest patching task and also exposes it as a property on the bugsnag plugin extension (with the current randomized behavior as the default).

Example:
```gradle
bugsnag {
  buildUuid.set(UUID.nameUUIDFromBytes("user-defined-seed".getBytes()))
}
```

Note: this uses `UUID` instead of String to ensure it's a valid UUID input. It's `Serializable` so it's compatible with Gradle configuration caching and just mapped to string under the hood.

## Changeset

New `buildUuid` property added to `BugsnagPluginExtension`, 

## Tests

Existing tests should catch regressions, default behavior is the same

### DIFF
--- a/src/main/kotlin/com/bugsnag/android/gradle/BugsnagPlugin.kt
+++ b/src/main/kotlin/com/bugsnag/android/gradle/BugsnagPlugin.kt
@@ -22,7 +22,6 @@ import org.gradle.api.file.FileCollection
 import org.gradle.api.file.RegularFile
 import org.gradle.api.provider.Provider
 import org.gradle.api.tasks.TaskProvider
-import java.util.UUID
 
 /**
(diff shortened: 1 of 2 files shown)


mattdyoung:
Thanks @ZacSweers! We'll review when we can...

ZacSweers:
Closing as this is far too old to pursue. I still think it's a good feature but we'll have to figure out our own solution if we need this

PR #291, opened 2020-09-07
[PLAT-4903] Update Maze-runner to V2

Cawllec:
## Goal

Updates Maze-runner to the latest version, with appropriate changes to the feature and fixture files.  This will allow us to bring proxy changes and other newer features from Maze-runner directly into this repo's tests. 

## Design

Bulk update of the fixtures, steps and scenarios to use Maze V2 steps and methods.

I'll outline a couple of places where changes have been made I'm not especially happy with, and will look into fixing either as a part of this PR or in a separate ticket.


Gemfile
```diff
@@ -1,3 +1,3 @@
 source "https://rubygems.org"
 
-gem "bugsnag-maze-runner", git: 'https://github.com/bugsnag/maze-runner', :branch => 'v1'
+gem "bugsnag-maze-runner", git: 'https://github.com/bugsnag/maze-runner', branch: 'fix-build-api-steps'
```
Cawllec, line 3: This needs to be changed back once https://github.com/bugsnag/maze-runner/pull/125 has been merged/released.  I'm going to address a couple of rubocop things there first.

features/abi_apk_splits.feature
```diff
@@ -3,73 +3,89 @@ Feature: Plugin integrated in project with ABI APK splits
 @skip_agp4_1_or_higher
 Scenario: ABI Splits project builds successfully
     When I build "abi_splits" using the "standard" bugsnag config
-    Then I should receive 16 requests
+    And I wait to receive 16 requests
```
Cawllec, line 6: Several features like this are heavily dependent on payload ordering.  I intend to address this in a separate PR targeting this branch before this branch is merged.

features/abi_apk_splits.feature
```diff
@@ -3,73 +3,89 @@ Feature: Plugin integrated in project with ABI APK splits
...
+    And the field "appId" for multipart request equals "com.bugsnag.android.example"
 
 @skip_agp4_1_or_higher
 Scenario: ABI Splits automatic upload disabled
     When I build "abi_splits" using the "all_disabled" bugsnag config
+    And I wait for 5 seconds
     Then I should receive no requests
 
 @skip_agp4_1_or_higher
 @skip_agp3_5
```
Cawllec, line 83: Several scenarios are being skipped currently due to the payload ordering issue.  That will be addressed in the PR mentioned above.

features/fixtures/app/gradlew
```diff
@@ -1,5 +1,21 @@
 #!/usr/bin/env sh
 
+#
```
Cawllec, line 3: @fractalwrench I'm assuming this shouldn't be checked in?

features/fixtures/config/bugsnag/all_disabled.gradle
```diff
@@ -1,8 +1,8 @@
 apply plugin: 'com.bugsnag.android.gradle'
 
 bugsnag {
-    endpoint = "http://localhost:${System.env.MOCK_API_PORT}"
-    releasesEndpoint = "http://localhost:${System.env.MOCK_API_PORT}"
+    endpoint = "http://localhost:9339"
```
Cawllec, line 4: I've hard coded the mock server port, as while we can technically change it there is no practical way or reason to.

features/steps/gradle_plugin_steps.rb
```diff
@@ -54,13 +54,10 @@
...
-  begin
-    steps %Q{
-      When I build "#{module_config}" using the "#{bugsnag_config}" bugsnag config
-    }
-    assert(false, "Expected script to fail with non-zero exit code")
-  rescue SystemExit
-  end
+  Runner.environment["MODULE_CONFIG"] = module_config
+  Runner.environment["BUGSNAG_CONFIG"] = bugsnag_config
+  _, exit_code = Runner.run_script("features/scripts/bundle_project_module.sh", blocking: true)
```
Cawllec, line 59: I'm not overly fond of this re-write, but can't think of a better way of getting the exit_code.  It's using the tools we've previously written, but just feels off.

features/fixtures/app/module/config/bugsnag/all_disabled.gradle
```diff
@@ -0,0 +1,6 @@
+project.afterEvaluate {
+    project.bugsnag.endpoint = "http://localhost:9339"
```
Cawllec, line 2: @fractalwrench Several of these configs seem to be additions rather than updates.  I'm not entirely sure where they've come from, but if you can confirm they aren't needed I'll remove them.

features/app_bundle.feature
```diff
@@ -3,71 +3,77 @@ Feature: Generating Android app bundles
 @skip_agp3_5
```
fractalwrench, line 3: These are flaky tests and will need their request order sorting (just flagging this where I see it as I know it'll be addressed in a separate PR)

features/app_bundle.feature
```diff
@@ -3,71 +3,77 @@ Feature: Generating Android app bundles
...
-    And the field "appId" for multipart request 1 equals "com.bugsnag.android.example.foo"
+    And the request is valid for the Android Mapping API
+    And the field "apiKey" for multipart request equals "TEST_API_KEY"
+    And the field "versionCode" for multipart request equals "1"
+    And the field "versionName" for multipart request equals "1.0"
+    And the field "appId" for multipart request equals "com.bugsnag.android.example.foo"
 
 Scenario: Auto upload disabled
     When I bundle "default_app" using the "all_disabled" bugsnag config
+    And I wait for 5 seconds
```
fractalwrench, line 78: Is there any way that we can avoid this wait?

features/disabled_bugsnag.feature
```diff
@@ -2,4 +2,5 @@ Feature: Bugsnag disabled
 
 Scenario: No requests are received when bugsnag is disabled
     When I build "default_app" using the "disabled_bugsnag" bugsnag config
-    Then I should receive 0 requests
+    And I wait for 5 seconds
```
fractalwrench, line 5: There are a couple of places which now have waits - why are they necessary?

features/fixtures/app/gradlew
fractalwrench: It's fine to check this in generally in JVM projects. The mazerunner tests update this file if the value of `GRADLE_WRAPPER_VERSION` is set however, which can mean files in the working tree change when the tests run.

features/fixtures/app/module/config/bugsnag/all_disabled.gradle
fractalwrench: As part of v5 I believe the directory containing these snippets was moved to `features/fixtures/app/config` - so these should be safe to remove

fractalwrench (commented):
Also noticed that a few scenarios are failing locally for me:

>cucumber features/app_bundle.feature:34 # Scenario: Bundling multiple flavors automatically
cucumber features/fail_on_upload_error.feature:3 # Scenario: Upload successfully with API key, mapping file, and correct endpoint
cucumber features/fail_on_upload_error.feature:8 # Scenario: No uploads or build failures when obfuscation is disabled
cucumber features/flavors.feature:3 # Scenario: Flavors automatic upload on build
cucumber features/ndk_app_agp400.feature:4 # Scenario: NDK apps send requests
cucumber features/ndk_app_agp400.feature:38 # Scenario: Custom projectRoot is added to payload
cucumber features/ndk_app_agp400.feature:66 # Scenario: Custom objdump location

features/app_bundle.feature
Cawllec: Not really, we could shorten it to 1-3 seconds, but we need to allow a bit of a wait

features/disabled_bugsnag.feature
Cawllec: They're needed to make sure we don't have a "late" payload coming through.  If we don't expect a payload, without a wait the likelihood of false positive passes increases greatly.

Gemfile
twometresteve: Any reason not to merge and release #125 now?

features/disabled_bugsnag.feature
twometresteve: I think we tend to think about this more when expecting 0 requests, but not 1 or more (where the same issue of receiving too many requests could happen).  I guess we just consider 0 to be a special case more worthy of waiting for?  We do already have a ticket to address tests with no positive outcome, perhaps by capturing logs instead, so we can probably tolerate this for now.  I did think that we could put some functionality into Maze Runner itself to wait in just the 0 request case, but that's probably compounding the issue.

fractalwrench: fair enough - I'm happy with adding a wait if it's necessary for the test and there are plans to rewrite tests which expect 0 requests anyway

Gemfile
twometresteve: But also, how do we want to track Maze Runner from BAGP when it's not using the Docker image?  Just pinning it to a release (tag) and not worrying about trying to track the latest at this stage?

features/steps/gradle_plugin_steps.rb
twometresteve: Return the exit code from the method that invoked it seems perfectly reasonable to me.

twometresteve (commented):
I've given this a static review having seen @fractalwrench 's comments, confident in this eagle eye.  I'm happy to merge this once he approves.

Gemfile
Cawllec: Pretty much.  We should pin it to a version or branch tag and only update when necessary.

fractalwrench (approved):

fractalwrench:
This will require #299 before merging but LGTM otherwise.

Issue #287, opened 2020-09-01
Cannot call Task.finalizedBy() after task has started execution


stephanedupont:
I am in the process of migrating from Crashlytics to Bugsnap for some Android apps.

Adding the bugsnap android gradle plugin in the project result in the following error when I run a debug build from Android Studio:

`What went wrong:
Could not create task ':app:processBugsnagGoogle-debugManifest'.
Cannot call Task.finalizedBy(Object...) on task ':completeRhythmTrainer:processGoogleDebugManifest' after task has started execution.`

- If I remove the bugsnap plugin, the error message disappears.
- If I build a release flavor and not a debug flavor, the error message disappears

xander-jones:
Hey @stephanedupont, thank you for reporting this bug. It's an issue that we're aware of, and we're looking to fix. We'll keep this thread updated with changes.

pbajurko:
For me it happens only when building from AS but not form gradle command. Still it's integration blocking issue. @fractalwrench when can we expect next plugin release with fix? Do we have any snapshot releases?

fractalwrench:
We believe this issue occurs for users who run their app in Debug mode in Android Studio and are using Android Gradle Plugin v4.0.

We are investigating this as a matter of priority, but we have encountered some issues in finding a solution that works for everyone. We'll post an update once we have a full solution but in the meantime, as you say, you can use the command line or you can upgrade to 4.1.0 of the Android Gradle Plugin (although this is currently in RC). Another workaround is to disable the plugin for the debug buildType (along with any other build variants that don't require deobfuscated stacktraces in Bugsnag):

```groovy
bugsnag {
    variantFilter { variant ->
        // disables plugin for all variants of the 'debug' buildType
        def name = variant.name.toLowerCase()
        if (name.contains("debug")) {
            enabled = false
        }
    }
}
```

pbajurko:
Well problem is that this plugin also delivers Bugsnag native libs required for compilation(It should be done by separate plugin or at least it should not depend on enabled flag). And dropping library modules support in v5 didn't make it easier :) So I would like to try your latest fix since it should be enough for us. 

Btw I also tried AS 4.0 with Bugsnag example app(AGP 3.6) and plugin 5.0.1 and issue is also there. So it seems that issue somehow related to AS only?

phillipsam:
[...]

PR #294, opened 2020-09-08
Fix task dependencies and order for BugsnagManifestUuidTask in AGP <4.1.0

fractalwrench:
## Goal

The `BugsnagManifestUuidTask` is used to write a UUID as a `meta-data` element to the AndroidManifest for AGP < 4.1.0. This changeset resolves two bugs in the task's configuration:

1. The `BugsnagManifestUuidTask` configure action mutates a task other than itself, which causes #287 
2. The `BugsnagManifestUuidTask` can run after the `processResources` task, which causes the UUID to be written to the manifest input file _after_ it has been copied to the APK/App bundle

These bugs are by their nature non-deterministic which partly explains why they were not observed during UAT of v5.

## Changeset

The PR is split into two commits which each fix an individual issue. Both issues are restricted to AGP < 4.1.0, as later versions of AGP offer the superior `android.onVariants` API which gets rid of a lot of complexity associated with editing the manifest.

### Configure action mutates a task other than itself

The plugin registers the `BugsnagManifestUuidTask` to support [task configuration avoidance](https://docs.gradle.org/current/userguide/task_configuration_avoidance.html#sec:how_does_it_work). This avoids adding the UUID task to a Gradle build when it is not necessary - for example, if a user is running unit tests the task is not required at all.

[The guidelines](https://docs.gradle.org/current/userguide/task_configuration_avoidance.html#sec:task_configuration_avoidance_general) suggest that when a task is configured, it should not mutate other tasks inside its configuration action. This guideline is broken by the plugin by adding a dependency on the `processManifest` task. The majority of the time this doesn't appear to have any consequence, but when running in a debug build via Android Studio, this results in the indeterminate behaviour seen in #287. 

The fix for this is to add a task dependency between `BugsnagManifestUuidTask` and `processManifest`, and additionally to enforce that `BugsnagManifestUuidTask` must run after `processManifest`. Because configuration avoidance requires for a [task to be looked up](https://docs.gradle.org/current/userguide/task_configuration_avoidance.html#sec:how_do_i_reference_a_task), a configure action has been added for `processManifest` which calls `get()` on the `BugsnagManifestUuidTask` and triggers its creation.

This approach is more performant than the original suggestion of simply calling `get()` in the plugin, because the UUID task should only be created when the `processManifest` task is executing - and in this scenario the bugsnag plugin should always manipulate the manifest.

This was tested by running the example app against a local artefact and confirming that running from AS does not fail the build.

### BugsnagManifestUuidTask can run after the processResources task

In a typical Android build the `processManifest` task produces a merged `AndroidManifest.xml`, and the `processResources` task copies this manifest to a location where it can be packaged into an APK/App Bundle artefact.

The bugsnag plugin adds a `BugsnagManifestUuidTask` task which writes a UUID into the `AndroidManifest.xml`. This needs to run after `processManifest` so that the `AndroidManifest.xml` file exists, and before the `processResources` task so that the updated file is copied to the APK.

The plugin correctly adds a [task ordering](https://docs.gradle.org/current/userguide/more_about_tasks.html#sec:ordering_tasks) so that `BugsnagManifestUuidTask` runs after `processManifest`, but does not do the equivalent for `processResources`. In a clean build where every task is rerun this results in the correct task ordering:

clean-build

However, if a build makes use of [up-to-date checks](https://docs.gradle.org/current/userguide/more_about_tasks.html#sec:up_to_date_checks) this can alter the order in which tasks are run so that `processResources` occurs before `BugsnagManifestUuidTask`. This causes a manifest without a build UUID to be copied to the final APK:

up-to-date

The fix for this is to configure `processResources` so that it must run after the `BugsnagManifestUuidTask`. This has been verified in the Android example app by running `./gradlew assembleRelease --rerun-tasks && ./gradlew assembleRelease` before and after the change, then checking whether the AndroidManifest contains the build UUID in Android Studio's [APK analyzer](https://developer.android.com/studio/build/apk-analyzer).

## Testing

These bugs have laid bare a few gaps in our automated test coverage. I would suggest we investigate adding additional test infastructure as part of future work:

1. Adding a step to mazerunner which extracts the APK/App Bundle archive and inspects whether the binary encoded `AndroidManifest.xml` contains the correct build UUID. We should consider adding this to most of the existing scenarios.
2. Adding some steps to mazerunner which build the APK/App Bundle using a build cache. Currently all builds happen from a clean slate which can miss problems with task ordering.

### DIFF
--- a/src/main/kotlin/com/bugsnag/android/gradle/BugsnagPlugin.kt
+++ b/src/main/kotlin/com/bugsnag/android/gradle/BugsnagPlugin.kt
@@ -279,7 +279,7 @@ class BugsnagPlugin : Plugin<Project> {
 
             // Enforces correct task ordering. The manifest can only be edited inbetween
             // when the merged manifest is generated (processManifestProvider) and when
-            // the merged manifest is copied for use in packaging the artefact (processResourcesProvider).
+            // the merged manifest is copied for use in packaging the artifact (processResourcesProvider).
             // This ensures BugsnagManifestUuidTask runs at the correct time for both tasks.
             //
             // https://docs.gradle.org/current/userguide/more_about_tasks.html#sec:ordering_tasks
--- a/src/main/kotlin/com/bugsnag/android/gradle/BugsnagPlugin.kt
+++ b/src/main/kotlin/com/bugsnag/android/gradle/BugsnagPlugin.kt
@@ -264,18 +264,24 @@ class BugsnagPlugin : Plugin<Project> {
             val taskName = computeManifestTaskNameFor(output.name)
             val manifestInfoOutputFile = project.computeManifestInfoOutputV1(output)
             val buildUuidProvider = project.newUuidProvider()
-            project.tasks.register(taskName, BugsnagManifestUuidTask::class.java) {
+
+            val manifestTask = project.tasks.register(taskName, BugsnagManifestUuidTask::class.java) {
                 it.buildUuid.set(buildUuidProvider)
                 it.variantOutput = output
                 it.variant = variant
                 it.manifestInfoProvider.set(manifestInfoOutputFile)
-                val processManifest = output.processManifestProvider.orNull
-
-                if (processManifest != null) {
-                    processManifest.finalizedBy(it)
-                    it.dependsOn(processManifest)
-                }
-            }.flatMap(BaseBugsnagManifestUuidTask::manifestInfoProvider)
+                it.dependsOn(output.processManifestProvider)
+            }
+            output.processManifestProvider.configure {
+                // Trigger eager configuration of the manifest task. This creates the task
+                // and ensures that it is configured whenever the manifest is processed
+                // and avoids mutating the task directly which should be avoided.
+                //
+                // https://docs.gradle.org/current/userguide/task_configuration_avoidance.html
+                // #sec:task_configuration_avoidance_general
+                manifestTask.get()
+            }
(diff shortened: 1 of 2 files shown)


imjoehaines (approved):
👍 can confirm both issues are fixed

src/main/kotlin/com/bugsnag/android/gradle/BugsnagPlugin.kt
```diff
@@ -267,18 +267,35 @@ class BugsnagPlugin : Plugin<Project> {
...
-            project.tasks.register(taskName, BugsnagManifestUuidTask::class.java) {
+
+            val manifestTask = project.tasks.register(taskName, BugsnagManifestUuidTask::class.java) {
                 it.buildUuid.set(buildUuidProvider)
                 it.variantOutput = output
                 it.variant = variant
                 it.manifestInfoProvider.set(manifestInfoOutputFile)
-                val processManifest = output.processManifestProvider.orNull
+                it.dependsOn(output.processManifestProvider)
+                it.mustRunAfter(output.processManifestProvider)
```
pbajurko, line 277: just curious, this line is not needed and doesn't change anything, right? 

> Gradle ensures that all task dependencies and ordering rules are honored when executing tasks, so that the task is executed after all of its dependencies

fractalwrench: Strictly speaking I don't believe it's required, but I think it's helpful making it very explicit here that the task must run after the `processManifest` task.

PR #295, opened 2020-09-09
Fix bugsnag.sharedObjectPaths not uploading SO files

fractalwrench:
## Goal

If `bugsnag.sharedObjectPaths` was set by a user this was not respected because the `BugsnagPlugin` did not add the directory to the search path used for the `BugsnagUploadNdkTask`. This meant SO files in custom locations were not uploaded.

This changeset adds any custom directories to the search path and adds an E2E scenario to cover this functionality. It also adds the missing implementation of `skip_agp3_6` which is used to suppress flaky tests.

Note that CI is saturated at the time of writing, and it's anticipated that there might be a few changes required to the NDK tests to get them passing due to different request ordering. It's not anticipated that any changes will be required for non-test code.


src/main/kotlin/com/bugsnag/android/gradle/BugsnagUploadNdkTask.kt
```diff
@@ -125,8 +125,8 @@ sealed class BugsnagUploadNdkTask(
     ): Collection<File> {
         val splitArch = variantOutput.getFilter(VariantOutput.FilterType.ABI)
         return searchDirectories.flatMap { findSharedObjectFiles(it, splitArch) }
+            .toSortedSet(compareBy { it.absolutePath })
             // sort SO files alphabetically by architecture for consistent request order
```
imjoehaines, line 129:
```suggestion
```

imjoehaines (approved):

fractalwrench:
I believe the previous logic used for sorting was ignoring multiple SO files as it created a sortedSet ordered by the ABI architecture's name, which isn't unique. The E2E scenarios have been updated to account for this - on AGP 4+ the Travis CI environment seems to include `liblog.so` so this has been added to the assertions. This seems to be down to Travis having a subtly different environment to local machines.

Still waiting on CI to pass as there may be further flakes.

fractalwrench:
@imjoehaines - just tagging for a quick re-review of alterations to the E2E tests now that AGP 4+ seems to be passing CI. LMK if you want to talk it over

imjoehaines (approved):
LGTM, the sorted set/number of requests change makes sense

Maybe we could write a maze runner step that would figure out how many requests _should_ be made (based on the number of SO files) to try to prevent the same thing happening again? Probably a future piece of work though

PR #275, opened 2020-08-21
Support proxies via OkHttp

ZacSweers:
Re-publish of #270 

Continuation of #200 but implemented in the new OkHttp stack. This also opportunistically fixes the BugsnagReleasesTask not using the same shared client

PR #298, opened 2020-09-10
Support mapping upload from Android Gradle Plugin through a proxy

fractalwrench:
## Goal

Adds support for uploading mapping files with the plugin through a proxy.

This changeset is based off #275 and will add automated test coverage.

## Changeset

- Altered `BugsnagHttpClientHelper` to read the [JVM networking properties](https://docs.oracle.com/javase/7/docs/api/java/net/doc-files/net-properties.html) when creating a new HTTP client
- If `proxyHost/proxyPort` are set, construct a new `Proxy` and set it on `OkHttpClient.Builder#proxy`
- If `proxyUser/proxyPass` are set, create an authenticator and set it on `OkHttpClient.Builder#proxyAuthenticator`

## Tests

Adds E2E scenarios that the plugin can route requests through a local HTTP proxy, with and without credentials.

The test fixture has supplied the proxy information through JVM arguments, rather than in the `gradle.properties`, as this simplifies testing and is functionally equivalent.

### DIFF
--- a/src/main/kotlin/com/bugsnag/android/gradle/internal/BugsnagHttpClientHelper.kt
+++ b/src/main/kotlin/com/bugsnag/android/gradle/internal/BugsnagHttpClientHelper.kt
@@ -1,14 +1,23 @@
 package com.bugsnag.android.gradle.internal
 
+import com.bugsnag.android.gradle.BugsnagPluginExtension
 import com.bugsnag.android.gradle.internal.BuildServiceBugsnagHttpClientHelper.Params
+import okhttp3.Authenticator
 import okhttp3.OkHttpClient
+import org.gradle.api.Project
 import org.gradle.api.provider.Property
 import org.gradle.api.provider.Provider
 import org.gradle.api.services.BuildService
 import org.gradle.api.services.BuildServiceParameters
+import java.net.InetSocketAddress
+import java.net.PasswordAuthentication
+import java.net.Proxy
 import java.time.Duration
 
-/** A simple API for providing a shared [OkHttpClient] instance for shared use in upload tasks. */
+/**
+ * A simple API for providing a shared [OkHttpClient] instance for shared use in upload tasks. This
+ * also handles common shared configuration such as timeouts, proxies, etc.
+ */
 interface BugsnagHttpClientHelper : AutoCloseable {
     val okHttpClient: OkHttpClient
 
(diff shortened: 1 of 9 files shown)


src/main/kotlin/com/bugsnag/android/gradle/internal/BugsnagHttpClientHelper.kt
```diff
@@ -40,14 +69,38 @@ class LegacyBugsnagHttpClientHelper(
...
+        val socketAddress = InetSocketAddress("localhost", port.toInt())
+        val proxy = Proxy(Proxy.Type.HTTP, socketAddress)
+        builder.proxy(proxy)
+    }
+
+    val user: String? = System.getProperty("http.proxyUser")
+    val pass: String? = System.getProperty("http.proxyPassword")
+
+    if (user != null && pass != null) { // set the default authenticator if credentials set
+        java.net.Authenticator.setDefault(ProxyAuthenticator("user", "password"))
```
imjoehaines, line 101: This works for the test but I'm not sure it will in practice 😀 

```suggestion
        java.net.Authenticator.setDefault(ProxyAuthenticator(user, password))
```

src/main/kotlin/com/bugsnag/android/gradle/internal/BugsnagHttpClientHelper.kt
```diff
@@ -40,14 +69,38 @@ class LegacyBugsnagHttpClientHelper(
...
+    configureHttpProxySettings(builder)
+    return builder.build()
+}
+
+private fun configureHttpProxySettings(builder: OkHttpClient.Builder) {
+    val host: String? = System.getProperty("http.proxyHost")
+    val port: String? = System.getProperty("http.proxyPort")
+
+    if (host != null && port != null) { // set the proxy host/port if set
+        val socketAddress = InetSocketAddress("localhost", port.toInt())
```
imjoehaines, line 92:
```suggestion
        val socketAddress = InetSocketAddress(host, port.toInt())
```

imjoehaines (changes requested):

src/main/kotlin/com/bugsnag/android/gradle/internal/BugsnagHttpClientHelper.kt
fractalwrench: Whoops :D good catch!

src/main/kotlin/com/bugsnag/android/gradle/internal/BugsnagHttpClientHelper.kt
```diff
@@ -40,14 +69,38 @@ class LegacyBugsnagHttpClientHelper(
...
+        val socketAddress = InetSocketAddress(host, port.toInt())
+        val proxy = Proxy(Proxy.Type.HTTP, socketAddress)
+        builder.proxy(proxy)
+    }
+
+    val user: String? = System.getProperty("http.proxyUser")
+    val pass: String? = System.getProperty("http.proxyPassword")
+
+    if (user != null && pass != null) { // set the default authenticator if credentials set
+        java.net.Authenticator.setDefault(ProxyAuthenticator(user, password))
```
imjoehaines, line 101: My bad, the variable is called `pass`!

```suggestion
        java.net.Authenticator.setDefault(ProxyAuthenticator(user, pass))
```

imjoehaines (approved):

PR #299, opened 2020-09-10
V2 Maze-runner: Fix issues with Unknown payload ordering 

Cawllec:
## Goal

Allows tests that can't guarantee the order in which payloads will arrive to work without a defined expected order.


features/flavor_apk_splits.feature
```diff
@@ -5,63 +5,28 @@ Scenario: Flavor Density Split project builds successfully
...
-    And the request is valid for the Android Mapping API
-    And the field "versionCode" for multipart request equals "1"
-    And the field "appId" for multipart request equals "com.bugsnag.android.example.foo"
-    And I discard the oldest request
-
-    And the request is valid for the Android Mapping API
-    And the field "versionCode" for multipart request equals "3"
-    And the field "appId" for multipart request equals "com.bugsnag.android.example.foo"
+    Then 6 requests are valid for the build API and match the following:
+      | appVersionCode |
```
fractalwrench, line 9: It would be good to also include the `appId` in this assertion

features/flavors.feature
```diff
@@ -4,30 +4,15 @@ Scenario: Flavors automatic upload on build
     When I build "flavors" using the "standard" bugsnag config
     And I wait to receive 4 requests
 
-    Then the request is valid for the Build API
-    And the payload field "appVersion" equals "1.0"
-    And the payload field "apiKey" equals "TEST_API_KEY"
-    And the payload field "appVersionCode" equals "1"
-    And I discard the oldest request
+    Then 2 requests are valid for the build API and match the following:
```
fractalwrench, line 7: This step is also missing the `appId` assertion

fractalwrench (approved):
LGTM

PR #301, opened 2020-09-16
Remove uploadDebugBuildMappings flag in favour of variantFilter

fractalwrench:
## Goal

Removes the undocumented `uploadDebugBuildMappings` flag in favour of the more flexible `variantFilter` API. By default the `variantFilter` will stop the plugin adding tasks to debug build variants. A user can also specify their own variantFilter to stop tasks being added for unwanted build variants, which improves build performance. 

## Changeset

- Removed `uploadDebugBuildMappings` flag
- Removed duplicate and unused files from `features/fixtures/app/module/config`
- Set default `Action` on `variantFilter` which disables any debug build variants

## Testing

Relied on existing E2E scenarios, particularly in `extension_properties.feature`.


src/test/kotlin/com/bugsnag/android/gradle/PluginExtensionTest.kt
```diff
@@ -4,6 +4,7 @@ import org.gradle.api.Project
 import org.gradle.testfixtures.ProjectBuilder
 import org.junit.Assert.assertEquals
 import org.junit.Assert.assertFalse
+import org.junit.Assert.assertNotNull
```
imjoehaines, line 7: This doesn't seem to be used

```suggestion
```

imjoehaines (approved):

PR #307, opened 2020-09-21
Support Unity shared object files in upload task

fractalwrench:
## Goal

Supports uploading shared object mapping files generated from `libunity.so` in the upload task. Note that this does not generate the files themselves and only alters the upload task so that they are sent to the correct endpoint.

## Changeset

- Renamed `BugsnagUploadNdkTask` to `BugsnagUploadSharedObjectTask` as this will be capable of uploading both Unity/NDK SO mapping files
- Added `UploadType` which alters the endpoint/payload of the shared upload request. If the SO file is from the NDK, it will be sent to `/so-symbol` with a key of `soSymbolFile`; otherwise it will be sent to `so-symbol-table` with a key of `soSymbolTable`

## Testing

- Fixed existing mazerunner assertions so that they check whether the payload body of a build/mapping/NDK mapping request is valid
- Added `requests are valid for the android Unity NDK mapping API and match the following:` step which will be used by Unity SO requests once the plugin is capable of generating these mapping files

### DIFF
--- a/features/steps/gradle_plugin_steps.rb
+++ b/features/steps/gradle_plugin_steps.rb
@@ -85,8 +85,8 @@ Then('{int} requests are valid for the android NDK mapping API and match the fol
   end
 end
 
-Then('{int} requests are valid for the android Unity NDK mapping API and match the following:') do |request_count, data_table|
-  requests = get_android_ndk_mapping_requests
+Then('{int} requests are valid for the android unity NDK mapping API and match the following:') do |request_count, data_table|
+  requests = get_android_unity_ndk_mapping_requests
   assert_equal(request_count, requests.length, 'Wrong number of NDK mapping API requests')
   RequestSetAssertions.assert_requests_match requests, data_table
 
@@ -118,7 +118,7 @@ def valid_android_ndk_mapping_api?(request_body)
   assert_not_nil(request_body['soSymbolFile'])
 end
 
-def valid_android_ndk_mapping_api?(request_body)
+def valid_android_unity_ndk_mapping_api?(request_body)
   valid_mapping_api?(request_body)
   assert_not_nil(request_body['soSymbolTable'])
 end
--- a/features/steps/gradle_plugin_steps.rb
+++ b/features/steps/gradle_plugin_steps.rb
@@ -59,18 +59,40 @@ Then('{int} requests are valid for the build API and match the following:') do |
   requests = get_build_requests
   assert_equal(request_count, requests.length, 'Wrong number of build API requests')
   RequestSetAssertions.assert_requests_match requests, data_table
+
+  requests.each do |request|
+    valid_build_api?(request[:body])
+  end
 end
 
 Then('{int} requests are valid for the android mapping API and match the following:') do |request_count, data_table|
   requests = get_android_mapping_requests
   assert_equal(request_count, requests.length, 'Wrong number of mapping API requests')
   RequestSetAssertions.assert_requests_match requests, data_table
+
+  requests.each do |request|
+    valid_android_mapping_api?(request[:body])
+  end
 end
 
 Then('{int} requests are valid for the android NDK mapping API and match the following:') do |request_count, data_table|
   requests = get_android_ndk_mapping_requests
   assert_equal(request_count, requests.length, 'Wrong number of NDK mapping API requests')
   RequestSetAssertions.assert_requests_match requests, data_table
+
+  requests.each do |request|
+    valid_android_ndk_mapping_api?(request[:body])
+  end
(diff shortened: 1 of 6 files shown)


features/steps/gradle_plugin_steps.rb
```diff
@@ -59,18 +59,40 @@
...
   requests = get_android_ndk_mapping_requests
   assert_equal(request_count, requests.length, 'Wrong number of NDK mapping API requests')
   RequestSetAssertions.assert_requests_match requests, data_table
+
+  requests.each do |request|
+    valid_android_ndk_mapping_api?(request[:body])
+  end
+end
+
+Then('{int} requests are valid for the android Unity NDK mapping API and match the following:') do |request_count, data_table|
```
twometresteve, line 88: Splitting hair here, but why lower `a` on android but upper `U` on Unity?

features/steps/gradle_plugin_steps.rb
```diff
@@ -59,18 +59,40 @@
...
+  requests = get_android_ndk_mapping_requests
```
twometresteve, line 89: Will this become `get_android_unity_ndk_mapping_requests` at some point?

features/steps/gradle_plugin_steps.rb
```diff
@@ -87,18 +109,22 @@ def valid_build_api?(request_body)
...
 end
 
-# TODO can avoid duplication
 def valid_android_ndk_mapping_api?(request_body)
-  assert_equal($api_key, request_body['apiKey'])
+  valid_mapping_api?(request_body)
   assert_not_nil(request_body['soSymbolFile'])
+end
+
+def valid_android_ndk_mapping_api?(request_body)
```
twometresteve, line 121: Seems to be exactly the same as above - copy and pasted with an edit to follow?

fractalwrench: Updated the method name.

features/steps/gradle_plugin_steps.rb
fractalwrench: updated to use lowercase `u` for consistency with existing steps

features/steps/gradle_plugin_steps.rb
fractalwrench: Yep - updated accordingly

src/main/kotlin/com/bugsnag/android/gradle/BugsnagUploadSharedObjectTask.kt
```diff
@@ -29,19 +29,24 @@ import javax.inject.Inject
...
-internal open class BugsnagUploadNdkTask @Inject constructor(
+internal open class BugsnagUploadSharedObjectTask @Inject constructor(
     objects: ObjectFactory,
     projectLayout: ProjectLayout
 ) : DefaultTask(), AndroidManifestInfoReceiver, BugsnagFileUploadTask {
 
+    enum class UploadType {
+        NDK,
+        UNITY
+    }
```
imjoehaines, line 40: Is it worth putting the endpoint and upload key as methods on this class, so the conditionals aren't needed later on? e.g.

```kotlin
enum class UploadType {
    NDK {
        override fun endpoint(base: String) = "${base}/so-symbol"
        override fun uploadKey() = "soSymbolFile"
    },

    UNITY {
        override fun endpoint(base: String) = "${base}/so-symbol-table"
        override fun uploadKey() = "soSymbolTableFile"
    };

    abstract fun endpoint(base: String): String
    abstract fun uploadKey(): String
}
```

then using it is just

```kotlin
val requestEndpoint = uploadType.get().endpoint(endpoint.get())
val soUploadKey = uploadType.get().uploadKey()
```

imjoehaines: Or even

```kotlin
enum class UploadType(private val path: String, val uploadKey: String) {
    NDK("so-symbol", "soSymbolFile"),
    UNITY("so-symbol-table", "soSymbolTableFile");

    fun endpoint(base: String): String {
        return "${base}/${path}"
    }
}

val requestEndpoint = uploadType.get().endpoint(endpoint.get())
val soUploadKey = uploadType.get().uploadKey
```

imjoehaines (approved):

src/main/kotlin/com/bugsnag/android/gradle/BugsnagUploadSharedObjectTask.kt
fractalwrench: good idea 👍  I've altered the enum accordingly

PR #310, opened 2020-09-23
Add basic unity E2E test fixtures

fractalwrench:
## Goal

Adds basic E2E test fixtures for Unity 2018.4 and Unity 2019.4 [exported Gradle projects](https://docs.unity3d.com/Manual/android-gradle-overview.html). These have been generated from the bugsnag-unity example project, obfuscation has been enabled, and the bugsnag gradle plugin has been applied to the project.

Currently this just ensures that the Unity projects upload JVM mapping files and release information. Separate changesets will be made to verify that SO files for NDK + Unity can be uploaded.

PR #311, opened 2020-09-24
Register unity shared object generation and upload tasks

fractalwrench:
## Goal

Registers Gradle tasks for generating SO mapping files from `libunity.so` and uploading the SO file. Note that this changeset simply registers the tasks and ensures that they run in a Unity project - the actual generation and verification that symbols are uploaded will take place in separate PRs.

## Changeset

- Changed SO mapping files from being output into the intermediate directory `soMappings` to two separate directories for NDK + Unity
- Added `BugsnagGenerateUnitySoMappingTask`, which mimics `BugsnagGenerateNdkSoMappingTask` but at present performs no action for its task
- Added a `BugsnagGenerateUnitySoMappingTask` and `BugsnagUploadUnityTask` for each build variant when `uploadNdkUnityLibraryMappings` is enabled

## Testing

Ran the Unity test fixtures added in #310 with `VERBOSE=true` to confirm that the tasks were run in Unity projects.

### DIFF
--- a/src/main/kotlin/com/bugsnag/android/gradle/BugsnagPlugin.kt
+++ b/src/main/kotlin/com/bugsnag/android/gradle/BugsnagPlugin.kt
@@ -18,6 +18,7 @@ import com.bugsnag.android.gradle.internal.register
 import com.bugsnag.android.gradle.internal.versionNumber
 import org.gradle.api.Plugin
 import org.gradle.api.Project
+import org.gradle.api.Task
 import org.gradle.api.file.FileCollection
 import org.gradle.api.file.RegularFile
 import org.gradle.api.provider.Provider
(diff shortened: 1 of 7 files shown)


src/main/kotlin/com/bugsnag/android/gradle/BugsnagPlugin.kt
```diff
@@ -366,6 +398,25 @@ class BugsnagPlugin : Plugin<Project> {
...
+        manifestInfoFileProvider: Provider<RegularFile>
+    ): TaskProvider<out BugsnagGenerateUnitySoMappingTask> {
+        // Create a Bugsnag task to upload Unity mapping file(s)
+        val outputName = taskNameForOutput(output)
+        val taskName = "generateBugsnagUnity${outputName}Mapping"
+        return BugsnagGenerateUnitySoMappingTask.register(project, taskName) {
+            variantOutput = output
+            objDumpPaths.set(bugsnag.objdumpPaths)
+            manifestInfoFile.set(manifestInfoFileProvider)
+            // TODO update me
```
fractalwrench, line 416: This will require additional inputs when the task actually generates mapping files, but for now requires no action

imjoehaines (approved):

PR #312, opened 2020-09-25
Generate shared object mapping files for libunity and libil2cpp

fractalwrench:
## Goal

Makes `BugsnagGenerateUnitySoMappingTask` generate a shared object mapping file for the `libunity` and `libil2cpp` SO files.

## Changeset

- `BugsnagGenerateUnitySoMappingTask` searches one directory above the project's root directory for Unity symbols. In an [exported Unity project](https://docs.unity3d.com/Manual/android-gradle-overview.html#building-or-exporting-a-gradle-project) the `libunity/libil2cpp` symbols are output in a gzip archive in this directory. (For Unity 2018 this is two directories above due to a different Gradle project structure)
- Searched the gzip archive for `libunity/libil2cpp` entries and copied them to `intermediates/bugsnag/unitySoFiles/<abi>/<so-file>`
- If no gzip archive is found, log a warning
- Added `SharedObjectType` to `SharedObjectMappingFileFactory`, which alters the objdump command used to generate mapping files
- Used `objdump --sym <so file>` to generate SO mapping files for Unity

### Future changes

Several changes are planned as part of this ticket but haven't been included in this PR for ease of review:

- Using grep to reduce SO mapping file size
- Filtering unnecessary archives by ABI
- Searching `Temp/StagingArea/libs` when using Unity's internal build system

## Testing

Added unit tests that verify methods which check whether an SO file/gzip entry are from Unity. Updated existing E2E tests to verify that a SO mapping file is uploaded for each available ABI of `libunity` and `libil2cpp`.

### DIFF
--- a/src/main/kotlin/com/bugsnag/android/gradle/BugsnagGenerateUnitySoMappingTask.kt
+++ b/src/main/kotlin/com/bugsnag/android/gradle/BugsnagGenerateUnitySoMappingTask.kt
@@ -4,6 +4,9 @@ import com.android.build.gradle.api.ApkVariantOutput
 import com.bugsnag.android.gradle.SharedObjectMappingFileFactory.UNITY_SO_MAPPING_DIR
 import com.bugsnag.android.gradle.internal.mapProperty
 import com.bugsnag.android.gradle.internal.register
+import okio.buffer
+import okio.sink
+import okio.source
 import org.gradle.api.DefaultTask
 import org.gradle.api.Project
 import org.gradle.api.file.DirectoryProperty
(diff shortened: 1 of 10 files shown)


src/main/kotlin/com/bugsnag/android/gradle/BugsnagGenerateUnitySoMappingTask.kt
```diff
@@ -48,19 +55,135 @@ internal open class BugsnagGenerateUnitySoMappingTask @Inject constructor(
...
+    val rootProjectDir: DirectoryProperty = objects.directoryProperty()
+
     @TaskAction
     fun generateMappingFiles() {
         logger.lifecycle("Generating Unity mapping files")
-        // TODO generate mapping file here
+
+        // TODO search Unity internal build too
+        val symbolArchives = getUnitySymbolArchives(rootProjectDir)
+        logger.info("Found symbol archives: $symbolArchives")
```
imjoehaines, line 71: This might be confusing if it's empty and so warns immediately afterwards — maybe it'd be better in an `else` to the if statement below?

imjoehaines (approved):

src/main/kotlin/com/bugsnag/android/gradle/BugsnagGenerateUnitySoMappingTask.kt
fractalwrench: Moved to an `else` statement

PR #313, opened 2020-09-29
Reduce size of Unity SO mapping files

fractalwrench:
## Goal

Reduces the size of Unity SO mapping files by filtering out unnecessary lines from the output. Specifically, this removes anything that is not a section header or function definition. Any symbols which are undefined are removed also.

This only applies to `libunity/libil2cpp` files as NDK SO files use a different format.

## Testing

Added a unit test to verify line filtering logic. I additionally verified that the SO file generated in the plugin matches the checksum of that generated from `objdump --sym <so file> | grep " F \|SYMBOL TABLE:" | grep -v ”*UND*”`, which confirms that only the unnecessary information has been removed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 * Generate shared object mapping files for libunity and libil2cpp
   [#312](https://github.com/bugsnag/bugsnag-android-gradle-plugin/pull/312)
+  [#313](https://github.com/bugsnag/bugsnag-android-gradle-plugin/pull/313)
 
 * Register unity shared object generation and upload tasks
   [#311](https://github.com/bugsnag/bugsnag-android-gradle-plugin/pull/311)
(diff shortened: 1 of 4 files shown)


imjoehaines (commented):
Changes look good, but is there a way we can test this with maze runner? Maybe just the fixture uploads a mapping file that's less than _x_ bytes?

A step like this would do that:

```ruby
request = Server.stored_requests.first # or a request number
body = JSON::dump(request[:body])

expected_maximum_size = 250_000 # whatever number of bytes makes sense
assert_operator(body.bytesize, :<, expected_maximum_size)
```

fractalwrench:
I’m not sure that E2E tests are the way to go here, my reasoning being:

1. Unnecessary lines being removed doesn’t affect the pipeline’s processing of the file and is only done to speed up uploads
2. the order in which the gradle plugin sends requests is indeterminate, which means it could be quite unwieldy to add a new mazerunner step that verifies this
3. the expected size of the uploaded gzip archive will vary between Unity versions/when updating our IO library, so will require updating whenever that happens
4. this wasn’t included in the original test plan on the design

imjoehaines (approved):

PR #316, opened 2020-09-30
Automatically search for NDK shared object files in unity projects

fractalwrench:
## Goal

NDK shared object mapping files are not currently uploaded by the plugin in Unity projects. This changeset ensures that mapping files are now uploaded automatically.

## Changeset

In addition to the user-defined value of `bugsnag.sharedObjectPaths`, the plugin will automatically search two additional directories: one which covers Unity projects using the Gradle build system, the other which covers Unity 2019+ projects which have been exported to a standalone Gradle project. SO mapping files will be generated for any SO files in these directories.

This additionally fixes a bug with the default behaviour of `isNdkUploadEnabled`, so that it returns true by default for Unity projects. Previously this heuristic checked whether cmake was in use in a project; however, in a Unity project this would not be the case.

## Testing

Adapted E2E tests so that NDK SO files are uploaded against a test fixture of an exported Unity project. Additionally verified in a non-exported Unity project that SO mapping files were generated when NDK files were present, and added unit tests for `bugsnag.sharedObjectPaths` behaviour.


src/main/kotlin/com/bugsnag/android/gradle/BugsnagPlugin.kt
```diff
@@ -566,13 +569,43 @@ class BugsnagPlugin : Plugin<Project> {
...
+    internal fun isNdkUploadEnabled(
+        bugsnag: BugsnagPluginExtension,
+        android: AppExtension
+    ): Boolean {
         val usesCmake = android.externalNativeBuild.cmake.path != null
         val usesNdkBuild = android.externalNativeBuild.ndkBuild.path != null
         val default = usesCmake || usesNdkBuild
-        return bugsnag.uploadNdkMappings.getOrElse(default)
+        val unityEnabled = isUnityLibraryUploadEnabled(bugsnag, android)
+        return bugsnag.uploadNdkMappings.getOrElse(default) || unityEnabled
```
imjoehaines, line 580: Is it intended for this to still be `true` in unity projects if `uploadNdkMappings` is `false`? I think this should be

```kotlin
val default = usesCmake || usesNdkBuild || isUnityLibraryUploadEnabled(bugsnag, android)
return bugsnag.uploadNdkMappings.getOrElse(default)
```

fractalwrench: Nice catch, it should indeed be this.

imjoehaines (approved):

PR #321, opened 2020-10-07
Remove filtering of Unity SO files

fractalwrench:
## Goal

Removes the filtering out of 'unnecessary' lines in Unity library SO files added in #313. This information will now be required by the pipeline so shouldn't be stripped from the file uploaded to Bugsnag.

## Testing

Relied on existing E2E coverage, also manually tested and verified that symbolication appears as expected.


src/main/kotlin/com/bugsnag/android/gradle/SharedObjectMappingFileFactory.kt
```diff
@@ -152,21 +130,15 @@ internal object SharedObjectMappingFileFactory {
...
-    ) {
+    private fun outputZipFile(stdout: InputStream, outputFile: File) {
         val src = stdout.source().buffer()
         val sink = outputFile.sink().gzip().buffer()
         src.use { source ->
             sink.use { gzipSink ->
                 // filter unnecessary lines from objdump here
                 generateSequence { source.readUtf8Line() }
-                    .filter(lineFilter)
                     .forEach { line -> gzipSink.writeString("$line\n", Charset.defaultCharset()) }
```
imjoehaines, line 142: Would [putting this back to using `writeAll`](https://github.com/bugsnag/bugsnag-android-gradle-plugin/blob/2e475db008a8f421b9f22f0807fa5786bd8313e3/src/main/kotlin/com/bugsnag/android/gradle/SharedObjectMappingFileFactory.kt#L138-L144) be more efficient than going line by line?

imjoehaines (approved):

src/main/kotlin/com/bugsnag/android/gradle/SharedObjectMappingFileFactory.kt
fractalwrench: Updated to use the old method.

PR #334, opened 2020-10-23
Automatically add maven repository for react-native AARs

fractalwrench:
## Goal

Automatically adds a [maven repository](https://docs.gradle.org/current/userguide/declaring_repositories.html#sec:declaring_multiple_repositories) for the directory in the `@bugsnag/react-native` node_module which contains the bugsnag-android AARs. This simplifies the setup for React Native apps using bugsnag, as they will no longer need to add [`bugsnag-react-native.gradle`](https://docs.bugsnag.com/platforms/react-native/react-native/#android) in their build scripts.

The repository is added for any project using react-native as it is required for the app to compile.

## Testing

Ran the example react-native app in the following scenarios:

1. current instructions: `apply from: "../../node_modules/@bugsnag/react-native/bugsnag-react-native.gradle"` 
2. no plugin or current instructions
3. the bugsnag plugin applied to the project

The build failed for 2 which confirms that the plugin picks up the AARs in the node module.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## TBD
 
+* Automatically add maven repository for react-native AARs
+  [#334](https://github.com/bugsnag/bugsnag-android-gradle-plugin/pull/334)
+
 * Populate task inputs for BugsnagUploadJsSourceMapTask
   [#332](https://github.com/bugsnag/bugsnag-android-gradle-plugin/pull/332)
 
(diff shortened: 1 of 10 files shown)


imjoehaines (commented):
Is this something we can add a MR test for?

fractalwrench:
@imjoehaines I've integrated bugsnag into the react-native test fixtures, so compilation of the E2E scenarios would now fail if the plugin couldn't pick up the AARs.

PR #337, opened 2020-10-30
Allow decompression of JVM mapping files in mazerunner

fractalwrench:
## Goal

Allows JVM mapping files to be inflated from a gzip archive in mazerunner. This adds a mazerunner step which allows the developer to assert that mapping entries exist in the mapping file for a list of symbol names.

Existing E2E scenarios have also been updated so that they verify the contents of the mapping files. This aims to get a representative number of scenarios, including:

- Default JVM app
- App with productFlavors
- App with APK splits
- NDK app


twometresteve (approved):
LGTM, subject to CI.

PR #350, opened 2020-12-08
Support AGP 4.2

fractalwrench:
## Goal

Adds support for 4.2 of the Android Gradle Plugin. This changeset alters the plugin to use a new API for manipulating the manifest when the AGP version is high enough. This is necessary as the previous API added in 4.1 was removed entirely. Further details of the API change can be found [here](https://android-developers.googleblog.com/2020/12/announcing-android-gradle-plugin.html)

## Changeset

- Extracted logic for registering manifest manipulation task into `ManifestUuidTaskV2Compat`
- Setup the plugin to compile groovy before kotlin, so that the plugin can call into Groovy compatibility shims whenever AGP breaks its API
- Updated BAGP to compile against AGP 4.2, and used the new `androidComponents.onVariants` API when it is available
- Projects using AGP 4.1 now call into the Groovy compatibility shim which can access the obsolete API
- Cherry picked a commit that temporarily disables failing NDK E2E tests due to a Travis CI environment change

## Testing

Ran E2E tests on AGP 4.2 on Travis CI.

Altered the `PluginExtension` unit test to use a mocked version of AGP. This is necessary as AGP 4.2 has started using [shared build services](https://docs.gradle.org/current/userguide/build_services.html#registering_a_build_service) which are difficult to setup in a test target.


.travis.yml
```diff
@@ -53,39 +53,39 @@ jobs:
...
-      script: bundle exec maze-runner -c --verbose
+      script: bundle exec maze-runner -c --verbose --fail-fast
 
       # AGP 4.2.0 E2E tests
     - name: AGP 4.2.0 E2E tests
       env:
-        - AGP_VERSION=4.2.0-alpha07
-        - GRADLE_WRAPPER_VERSION=6.6-rc-6
-      script: bundle exec maze-runner -c --verbose
+        - AGP_VERSION=4.2.0-beta03
```
imjoehaines, line 89: If this was already running on Travis and was passing, does that mean we don't have tests for the changes in this PR?

fractalwrench: Good question - the breaking change was introduced in AGP 4.2.0-alpha14, - so alpha07 would pass CI fine using the old API.

imjoehaines (approved):

PR #352, opened 2020-12-09
Append sourcemap path to endpoint provided by plugin

fractalwrench:
## Goal

The plugin needs to add the ~`/sourcemap`~ `/react-native-source-map` route when a custom endpoint is provided, which is then passed to the source map CLI.

## Changeset

If the endpoint is `upload.bugsnag.com`, ~just pass that~ pass nothing. Otherwise, ~`sourcemap`~ `react-native-source-map` should be appended onto the route.

The version of bugsnag-sourcemaps was also bumped to the latest available from npm.

## Testing

Added to existing unit tests to verify that the endpoint is calculated correctly.

### DIFF
--- a/src/main/kotlin/com/bugsnag/android/gradle/BugsnagPlugin.kt
+++ b/src/main/kotlin/com/bugsnag/android/gradle/BugsnagPlugin.kt
@@ -40,6 +40,7 @@ import org.gradle.api.file.RegularFile
 import org.gradle.api.provider.Provider
 import org.gradle.api.tasks.TaskProvider
 import java.io.File
+import java.net.URL
 import java.util.UUID
 
 /**
@@ -482,7 +483,7 @@ class BugsnagPlugin : Plugin<Project> {
             bundleJsFileProvider.set(File(rnBundle))
             sourceMapFileProvider.set(File(rnSourceMap))
             overwrite.set(bugsnag.overwrite)
-            endpoint.set(bugsnag.endpoint)
+            endpoint.set(getSourcemapUploadEndpoint(bugsnag))
             devEnabled.set("true" == dev)
             failOnUploadError.set(bugsnag.failOnUploadError)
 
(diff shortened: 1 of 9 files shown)


src/main/kotlin/com/bugsnag/android/gradle/BugsnagPlugin.kt
```diff
@@ -497,6 +498,16 @@ class BugsnagPlugin : Plugin<Project> {
         }
     }
 
+    internal fun getSourcemapUploadEndpoint(bugsnag: BugsnagPluginExtension): String {
+        return when (val endpoint = bugsnag.endpoint.get()) {
+            UPLOAD_ENDPOINT_DEFAULT -> endpoint
```
bengourley, line 503: This looks like it passes through the endpoint (host only) if it is set, or left as the default value?

`@bugsnag/source-maps` needs the absolute URL to be passed (including the path). So you can either construct the full path, or don't pass any value and it will use the default. As it currently stands this will try to send a react native source map payload to the root of the upload server and will fail, because the payloads are different.

(Note that the path should be `react-native-source-map` per my other comment.

src/main/kotlin/com/bugsnag/android/gradle/BugsnagPlugin.kt
```diff
@@ -497,6 +498,16 @@ class BugsnagPlugin : Plugin<Project> {
...
+            else -> {
+                val host = URL(endpoint)
+                return URL(host, "sourcemap").toString()
```
bengourley, line 506: My bad, I told you it was this. It should actually be the RN path:

```suggestion
                return URL(host, "react-native-source-map").toString()
```

bengourley (changes requested):

src/main/kotlin/com/bugsnag/android/gradle/BugsnagPlugin.kt
fractalwrench: Updated so that the default endpoint is not passed as part of the command shelled out. It still remains as a property on the task as this is important for Gradle's task caching.

bengourley (approved):

PR #359, opened 2021-01-19
Add support for DexGuard 9

fractalwrench:
## Goal

Adds support for uploading mapping files generated by DexGuard 9, which has changed the location where mapping files are output.

## Changeset

- Retrieved the DexGuard version from `dexguard.version` or `dexguard.path`, one of which is always set. This is used to change the behaviour depending on whether v9 is in use or not.
- If DexGuard <=8 is in use, read the mapping file from `build/outputs/mapping`, otherwise look in the new location of `build/outputs/dexguard/mapping`
- If DexGuard is in use, ensure that the BAGP task always runs after the dexguard task, so that the mapping file has been generated and is ready for upload. DexGuard generates the mapping file later in the build process than AGP as it reprocesses the APK generated by AGP.
- `BugsnagGenerateProguardTask` accepts multiple mapping files as a [task input](https://docs.gradle.org/current/userguide/more_about_tasks.html#sec:up_to_date_checks) because the mapping file location alters for DexGuard depending on whether an APK or App Bundle is being generated. The task action resolves which file exists when it is run and uploads that

## Testing

Added unit test coverage for DexGuard mapping file calculations, and relied on E2E tests to confirm there were no regressions for regular AGP users.

As DexGuard is not something that we're able to create public tests for, functionality was manually tested by confirming that mapping files were uploaded for the following scenarios on both DexGuard 8 & 9:

- A universal APK
- An app bundle
- An APK with productFlavors
- An App Bundle with productFlavors
- An APK with ABI splits
- An APK with productFlavors and ABI splits

### DIFF
--- a/src/main/kotlin/com/bugsnag/android/gradle/internal/DexguardCompat.kt
+++ b/src/main/kotlin/com/bugsnag/android/gradle/internal/DexguardCompat.kt
@@ -1,0 +1,86 @@
+package com.bugsnag.android.gradle.internal
+
+import com.android.build.gradle.api.ApkVariant
+import com.android.build.gradle.api.ApkVariantOutput
+import com.bugsnag.android.gradle.GroovyCompat
+import org.gradle.api.Project
+import org.gradle.util.VersionNumber
+import java.io.File
+import java.nio.file.Paths
+
+/**
+ * Finds the mapping file locations for DexGuard >=9. This can be different depending on whether it
+ * is a bundle or an APK.
+ */
+internal fun findMappingFileDexguard9(
+    project: Project,
+    variant: ApkVariant,
+    variantOutput: ApkVariantOutput
+): List<File> {
+    return listOf(
+        findDexguardMappingFile(project, variant, variantOutput, "outputs", "dexguard", "mapping", "apk"),
+        findDexguardMappingFile(project, variant, variantOutput, "outputs", "dexguard", "mapping", "bundle")
+    )
+}
+
+/**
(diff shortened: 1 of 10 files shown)


fractalwrench:
@twometresteve tagged you to make sure you're happy with the test approach on this one

src/main/kotlin/com/bugsnag/android/gradle/internal/DexguardCompat.kt
```diff
@@ -0,0 +1,85 @@
...
+internal fun Project.hasDexguardPlugin(): Boolean {
+    return pluginManager.hasPlugin("dexguard")
+}
+
+/**
+ * Retrieves the major version of DexGuard in use in the project
+ */
+internal fun getDexguardMajorVersionInt(project: Project): Int {
+    val version = GroovyCompat.getDexguardVersionString(project) ?: return -1
+    val major = version[0]
```
imjoehaines, line 74: Won't this break when DexGuard 10 is released? It might be safer to use `VersionNumber` to parse this, like we do elsewhere

imjoehaines (approved):

src/main/kotlin/com/bugsnag/android/gradle/internal/DexguardCompat.kt
fractalwrench: Good call - I've updated the code to use `VersionNumber` for parsing this.

twometresteve (commented):
Test approach seems reasonable to me given the constraints.

PR #362, opened 2021-01-28
Support uploading JS sourcemaps generated with Hermes

fractalwrench:
## Goal

Supports uploading JS sourcemaps generated with Hermes. Previously the plugin introspected react.gradle's `--sourcemap-output`, but this is the wrong file for Hermes apps. The sourcemap output location is therefore calculated manually instead by mimicking [RN's logic](https://github.com/facebook/react-native/blob/master/react.gradle#L116).

## Changeset

Manually calculated the RN sourcemap file location for the JS sourcemap upload task.

## Testing

Relied on existing E2E test coverage to confirm that a sourcemap was uploaded.

Performed additional manual testing in an example app, confirming that the build was successful with:

- `uploadReactNativeMappings` disabled
- Hermes enabled
- Hermes disabled

It was observed that the same sourcemap path was passed to the RN CLI both when hermes was enabled/disabled.

It may be worth performing some E2E testing that symbolication on the dashboard works, but I'm not 100% sure on how to best go about that?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## TBD
+
+* Support uploading JS sourcemaps generated with Hermes
+  [#362](https://github.com/bugsnag/bugsnag-android-gradle-plugin/pull/362)
+
 ## 5.7.2 (2021-01-27)
 
 * Add support for DexGuard 9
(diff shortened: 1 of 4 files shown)


imjoehaines (approved):

bengourley (approved):
Ran a full manual end to end test. The correct source map is now uploaded for a Hermes build and symbolicates correctly on the dashboard.

PR #365, opened 2021-02-10
More robust DexGuard support

EmielVandeloo:
## Goal

Thank you for your efforts to ensure compatibility with Guardsquare's DexGuard and your recent additions for DexGuard version 9. However, we have received a couple of support tickets notifying us that in some cases, the upload of mapping files does not happen as expected. I would like to propose a couple of changes to achieve full compatibility with DexGuard.

## Design

First, we have noticed some problems when only screen density splits are enabled. They are treated specially in `DexGuardCompat.kt`, which shouldn't be the case. Since DexGuard 9.0.6, we have implemented full support for splits (i.e. we produce a mapping file in its own location for each split). When dealing with bundles, we have only one mapping file, for which no split directories should be taken into account.

Second, the approach of deciding which DexGuard version is being used (based on the path name containing `DexGuard-x.x.x`) is not very robust. I have changed it to using the manifest of the `dexguard.jar`, located in the `lib` directory of a DexGuard distribution. This is also the way we check for the version in our Gradle plugin and is more robust for users who rename directories.

## Changeset

- No special handling of density splits in `DexguardCompat.kt`.
- Appropriate handling of bundles in `DexguardCompat.kt`.
- Get the version from the manifest of `dexguard.jar` in `GroovyCompat.groovy`.
- Removed tests for getting the version from the path in `DexguardCompatKtTest.kt`.

## Testing

Manual testing with Bugsnag integrated in one of our samples. The stacktraces are correctly deobfuscated for obfuscated APK's with and without splits, and for obfuscated bundles.

### DIFF
--- a/src/main/groovy/com/bugsnag/android/gradle/GroovyCompat.groovy
+++ b/src/main/groovy/com/bugsnag/android/gradle/GroovyCompat.groovy
@@ -6,6 +6,8 @@ import org.gradle.api.Action
 import org.gradle.api.Project
 
 import java.nio.file.Paths
+import java.util.jar.Attributes
+import java.util.jar.Manifest
 
 /**
  * Contains functions which exploit Groovy's metaprogramming to provide backwards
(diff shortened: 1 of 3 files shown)


yousif-bugsnag:
Thanks @EmielVandeloo! We'll look to review this sometime soon.

ingdas:
Hi @yousif-bugsnag , some of our mutual customers are affected by this bug. Is there anything that we can do to help you move this forward?

xander-jones:
Hi @ingdas, thank you for your offer of support. This PR is on our list of to-do's, but we have a number of other tasks competing for priority at the moment. I suspect it will need some automated testing adding in as a few tests have been removed (because it was testing the previous path build method), we are open to further changes off the back of this one if there are some suitable tests you can think of adding in. But just to confirm, it's on our list to do 😊

fractalwrench:
Thanks for the contribution @EmielVandeloo - these changes are really helpful. I plan on merging the changes down as part of #377 so I'll close this PR.

PR #366, opened 2021-02-11
Move addition of the local Maven repo for ReactNative into afterEvaluate

imjoehaines:
## Goal

The `withPlugin` block runs before configuration has been applied. This means that a custom `nodeModulesDir` won't be present yet and so the default will always be used. Moving this to `afterEvaluate` ensures that we can use a custom `nodeModulesDir` if one has been given

Note that this happens before the `enabled` check so that the Bugsnag React Native package can still be resolved. Otherwise the build fails when BAGP is loaded but disabled

## Testing

Tested in a local monorepo setup with hoisting

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## TBD
+
+* Fix `nodeModulesDir` option not being applied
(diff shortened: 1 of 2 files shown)


fractalwrench (approved):
LGTM, pending the completion of CI & a changelog entry

PR #371, opened 2021-03-11
Add versionCode property to AndroidManifestInfoReceiver

ZacSweers:
## Goal

Right now the manifest is determined to be the source of truth for the application `versionCode`, but there are new APIs in AGP that allow setting it via `onVariants...outputs.versionCode`. This appears to be set _after_ manifest processing, so it results in uploaded mapping files that don't match on Bugsnag's backend for retracing due to a version code mismatch.

## Design

I've wired in just enough right now to set it manually, as a toe-hold for the future. It adds an optional `versionCode` property that will take precedence if set but otherwise defer to the manifest version (which matches AGP's behavior). Integrating this across the many AGP versions that the Bugsnag plugin supports will take more time. I'd like to propose going with this as a toe-hold for now and consumers like us can manually set it as-needed while a longer term integration is sought, as it would impact the current way variants are loaded and require newer APIs.

## Testing

Just a unit test for now, planning to test this in our project to make sure the versionCode override works

### DIFF
--- a/src/main/kotlin/com/bugsnag/android/gradle/AndroidManifestInfo.kt
+++ b/src/main/kotlin/com/bugsnag/android/gradle/AndroidManifestInfo.kt
@@ -24,9 +24,14 @@ data class AndroidManifestInfo(
     internal companion object {
         private val ADAPTER = Moshi.Builder().build().adapter(AndroidManifestInfo::class.java)
 
-        fun read(file: File): AndroidManifestInfo {
-            return file.source().buffer().use {
+        fun read(file: File, versionCode: Int?): AndroidManifestInfo {
+            val info = file.source().buffer().use {
                 ADAPTER.fromJson(it) ?: error("Failed to parse manifest info.")
+            }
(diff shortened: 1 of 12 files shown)


ZacSweers:
The last two commits were opportunistic to make it easy to filter on these types when using `tasks.withType<...>()` APIs to filter

johnkiely1:
Hey @ZacSweers, 

Thanks for this, we will review as soon as we can.

ZacSweers:
Confirmed internally that using this change and setting the version code property resolves the original issue for us

fractalwrench:
Thanks for the contribution @ZacSweers - these commits will be merged with #375 so I'm closing this PR.

xander-jones:
Hey @ZacSweers, this got merged down through https://github.com/bugsnag/bugsnag-android-gradle-plugin/pull/375 and was released in https://github.com/bugsnag/bugsnag-android-gradle-plugin/releases/tag/v5.7.5. Thank you for your contribution!

PR #381, opened 2021-04-12
Avoid release build failure for invalid API key

IslamSalah:
## Goal

On integrating `bugsnag-android-gradle-plugin` to upload mappings at the build time, our release builds fail on having invalid api key. Normally the valid api key is fed to a placeholder so it can be accessible only in specific setups but invalid in others.

## Design

Ths proposed change should prevent failing release builds in setups where the api key shouldn't be accessible. It increases the flexibility of the plugin usage regardless having a valid/invalid api-key while at the same time retain logging errors if any. 

## Changeset

Similar to the case now in `bugsnag-android`, where an invalid api-key, [example](https://gist.github.com/IslamSalah/bab00b32e40ad4126c0b10b0636a0cc7), prevent sending events to the dashboard without failing builds. The proposed change stops concerned exceptions

## Testing

Automated tests should pass normally. Below screenshots might give more visibility 

#### Before change
before

#### After change
after

### DIFF
--- a/features/fail_on_upload_error.feature
+++ b/features/fail_on_upload_error.feature
@@ -20,3 +20,4 @@ Scenario: Upload failure due to connectivity failure
     When I build the failing "default_app" using the "wrong_endpoint" bugsnag config
     And I wait for 3 seconds
     Then I should receive no requests
+    Then the exit code equals 0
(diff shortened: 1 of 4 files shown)


IslamSalah:
On trying to fix the tests, I found the feature already implemented
failOnUploadError

PR #385, opened 2021-04-30
License-audit: Add dependency license auditing to buildkite pipeline

Cawllec:
## Goal
Adds license-auditing to buildkite pipeline.


twometresteve (commented):
LGTM as long as @fractalwrench is happy with the `biuld.gradle` changes.

fractalwrench (approved):
LGTM. `compile` will eventually need switching back to `implementation` when there's a push to move to Gradle 7, and at that point we'll either need to fix the license plugin or look at alternatives - but it feels ok to cross that bridge when we come to it.

PR #386, opened 2021-05-28
fix(gradle): replaced jcenter with mavenCentral

lemnik:
## Goal
Remove jcenter from the repository list of the bugsnag-android-gradle-project:

* Replace jcenter reference in the plugin project with mavenCentral
* Bump detekt to version 1.17.1 to avoid dependencies served only from jcenter
* Rerun the detekt baseline to cover the existing codebase

This does *not* cover the jcenter references in the test fixtures as these specifically cover versions of AGP that have transitive dependencies on jcenter.

## Testing
Relied on existing tests

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,7 +1,7 @@
 buildscript {
     repositories {
         maven { url "https://plugins.gradle.org/m2/" }
-        jcenter()
+        mavenCentral()
         google()
     }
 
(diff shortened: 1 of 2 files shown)


fractalwrench (approved):

fractalwrench:
@lemnik one note - this should target `next` rather than `master`

PR #387, opened 2021-05-28
Update mazerunner to v3.7.4

fractalwrench:
## Goal

Bumps mazerunner to the latest v3, which should reduce the amount of change in an eventual upgrade to v5.

## Testing

Relied on existing test coverage.


twometresteve (approved):
LGTM

PR #390, opened 2021-06-02
Emit a compatibility error when AGP 7+ is used

lemnik:
## Goal
Emit a helpful compatibility error when the plugin is used with Android Gradle Plugin 7 or higher, as this version of the Bugsnag Gradle Plugin is not compatible.

## Testing
A new test fixture has been included to ensure that AGP 7 inclusion fails the build.

### DIFF
--- a/src/main/kotlin/com/bugsnag/android/gradle/BugsnagPlugin.kt
+++ b/src/main/kotlin/com/bugsnag/android/gradle/BugsnagPlugin.kt
@@ -71,7 +71,7 @@ class BugsnagPlugin : Plugin<Project> {
         if (AgpVersions.CURRENT >= AgpVersions.VERSION_7_0) {
             throw StopExecutionException(
                 "Using com.bugsnag.android.gradle with Android Gradle Plugin 7+ " +
-                    "requires an upgrade to com.bugsnag.android.gradle:7.0.0. " +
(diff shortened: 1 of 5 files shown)


src/main/kotlin/com/bugsnag/android/gradle/BugsnagPlugin.kt
```diff
@@ -66,12 +68,22 @@ class BugsnagPlugin : Plugin<Project> {
 
     @Suppress("LongMethod")
     override fun apply(project: Project) {
+        if (AgpVersions.CURRENT >= AgpVersions.VERSION_7_0) {
+            throw StopExecutionException(
+                "Using com.bugsnag.android.gradle with Android Gradle Plugin 7+ " +
+                    "requires an upgrade to com.bugsnag.android.gradle:7.0.0. " +
+                    "For more information about this change, see " +
+                    "https://docs.bugsnag.com/platforms/android/"
```
fractalwrench, line 76: nit: the link will probably end up here instead: https://docs.bugsnag.com/build-integrations/gradle/

fractalwrench (approved):
LGTM pending resolution of the review comment.

I think the PR should also target `next`, and should avoid merging down until v7 is ready to release. If we end up creating additional changes to v5 then it be worth creating a separate v5 integration branch.

PR #393, opened 2021-06-03
Make the plugin compatible with AGP7

lemnik:
## Goal
Change the bugsnag-android-gradle-plugin to be compatible with Android Gradle Plugin 7. The plugin is now compiled against the AGP 7 API, all backwards compatible code has been removed.

## Testing
The maze-runner tests are not expected to pass yet, although they will now build and can run if run using AGP 7 and Gradle 7.


fractalwrench (commented):
I ran into some trouble running the tests locally, and also getting things to compile. Have the publish plugin steps changed? The docs might require a small tweak:

>Build file 'bugsnag-android-gradle-plugin/build.gradle' line: 24

>* What went wrong:
>An exception occurred applying plugin request [id: 'com.vanniktech.maven.publish', version: '0.15.1']
> Failed to apply plugin 'com.vanniktech.maven.publish'.
   > The property SONATYPE_NEXUS_USERNAME was removed. Use mavenCentralUsername instead.

Generally this looks good - it'd be nice to run the tests locally to confirm what is working and what isn't. Another thing that might be worth changing is altering the Travis CI matrix so that it only runs against AGP 7/Gradle 7 for now - as that should speed up the feedback loop and indicate what tests are passing/failing without running locally.

PR #394, opened 2021-06-07
Make the plugin compatible with AGP7

lemnik:
## Goal
Change the bugsnag-android-gradle-plugin to be compatible with Android Gradle Plugin 7. The plugin is now compiled against the AGP 7 API, all backwards compatible code has been removed.

## Testing
The maze-runner tests are not expected to pass yet, although they will now build and can run if run using AGP 7 and Gradle 7.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -5,23 +5,24 @@ buildscript {
         google()
     }
 
-    ext.agpVersion = "4.2.0-beta03" // compile against 4.1, but maintain compat to 3.4.0
-    ext.kotlinVersion = "1.3.72"
+    ext.agpVersion = "7.0.0-beta03"
+    ext.kotlinVersion = "1.4.31"
 
     dependencies {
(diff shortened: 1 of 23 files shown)


build.gradle
```diff
@@ -5,23 +5,24 @@ buildscript {
...
 plugins {
     id "java-gradle-plugin"
     id "groovy"
-    id "com.gradle.plugin-publish" version "0.12.0"
-    id "com.vanniktech.maven.publish" version "0.12.0"
-    id "com.github.hierynomus.license" version "0.15.0"
+    id 'maven-publish'
+    id "com.gradle.plugin-publish" version "0.15.0"
+    id "com.vanniktech.maven.publish" version "0.15.1"
+    //id "com.github.hierynomus.license" version "0.15.0"
```
lemnik, line 25: This plugin will need replacing or a patch to work with Gradle 7

fractalwrench (approved):
LGTM pending CI

PR #395, opened 2021-06-08
Only run test features on AGP7 / Gradle 7

lemnik:
## Goal
Avoid running test features on earlier versions of AGP / Gradle as they won't pass due to the compatibility checks.

This PR also disables all of our NDK tests due to a change in where AGP places it's intermediate `so` files. These tests will be re-enabled once we have a fix in-place for that change. 

## Testing
Testing relies on existing tests to illustrate the plugin compatibility. The only changes are to comment-out the NDK related tests (as mentioned above).


features/proxy.feature
```diff
@@ -39,51 +39,27 @@ Scenario: Authenticated HTTP proxy without creds
     Then I wait for 5 seconds
     And I should receive no requests
 
-@skip_agp4_0_or_higher
```
fractalwrench, line 42: These tags can probably be removed throughout the tests and where they're defined in the Ruby file.

.travis.yml
```diff
@@ -51,50 +55,8 @@ jobs:
...
-        - GRADLE_WRAPPER_VERSION=6.5.1
-        - RN_FIXTURE_DIR=features/fixtures/rn063/android
-      script: bundle exec maze-runner -c --verbose --fail-fast
-
-      # AGP 4.2.0 E2E tests
-    - name: AGP 4.2.0 E2E tests
+      # AGP E2E tests
+    - name: AGP E2E tests
       env:
-        - AGP_VERSION=4.2.0-beta03
```
fractalwrench, line 97: It might be nice to define these explicitly just so that it's obvious what is being used, and to make it easy to change in future from one place.

.travis.yml
```diff
@@ -51,50 +55,8 @@ jobs:
     - name: plugin install
       script: ./gradlew install
```
fractalwrench, line 56: This is an existing problem, but I think it might be worth tweaking this step so that it's combined with the unit test job, as it'll skip 9 minutes of what is mostly environment setup on CI.

fractalwrench (commented):
These changes look good, I left a few comments inline.

I think we will need to fix the license checks before merging this work - so it'd be good to get an idea of exactly how it's failing and what would be required to get it compatible with Gradle 7. Depending on the complexity we could look into forking it ourselves, finding a similar plugin, or rolling our own license check solution.

.travis.yml
lemnik: We do also (sort of) have this in `env.rb`: https://github.com/bugsnag/bugsnag-android-gradle-plugin/pull/395/files#diff-3c1d97c5f1fa7ca90a4a96df18478ce19fe4d3a8dbaa3914ccc6ec6c43e81cbe

fractalwrench (approved):
LGTM pending CI

PR #396, opened 2021-06-09
Emit compatibility error and warning

lemnik:
## Goal
When using bugsnag-android-gradle-plugin:7 with any version of the Android Gradle Plugin other than 7 we emit either an error or warning regarding the version compatibility. If the AGP version is less-than 7 we emit an error, while future major releases will trigger a warning that there may be incompatibilities.

## Testing
New automated test features are added testing all previously tested versions of AGP to ensure a build-failure.


fractalwrench (approved):
LGTM pending CI

PR #398, opened 2021-06-11
Work around the new shared-object location in AGP7

lemnik:
## Goal
Work around the new shared-object location in AGP7 by looking at the parent-of-parent (`$soFolder/../..`) directory and using the existing `so` file scanning to find all of the appropriate files.

The implementation will revert to the old approach if the parent-of-parent-of-parent (`$soFolder/../../..`) is not named `cxx` as a sanity check.

## Testing
Re-enabled the NDK test features, and modified the NDK fixture app to not include the `liblog.so` on Travis.

The Unity tests have also been enabled, and the fixture build scripts have been modified to work with AGP 7 (as this is a trivial and appears to be a reasonable change to make).

### DIFF
--- a/features/fixtures/unity_2018/example/build.gradle
+++ b/features/fixtures/unity_2018/example/build.gradle
@@ -67,13 +67,15 @@ android {
     buildTypes {
         debug {
             minifyEnabled false
-//            useProguard false
+            // useProguard is not compatible with Android Gradle Plugin 7+
+            // useProguard false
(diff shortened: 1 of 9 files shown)


features/fixtures/unity_2018/example/build.gradle
```diff
@@ -79,13 +67,13 @@ android {
     buildTypes {
         debug {
             minifyEnabled false
-            useProguard false
+//            useProguard false
```
fractalwrench, line 70: Could this be removed rather than commented out from the fixtures?

PR #404, opened 2021-06-28
Rescue the JS bundle from being overwritten by Hermes

lemnik:
## Goal
Upload the JS bundle instead of the Hermes bundle - when Hermes is enabled.

## Design
The Hermes stage of the `react.gradle` script is part of the standard ReactNative bundle task, but is in a `doLast` block (which produces an additional `Action` in the task). When Hermes is enabled for the project this change inserts a "rescue" action before the first `doLast` action it finds in the task. The rescue action copies the JS source bundle to the `intermediates/bugsnag/rescue` directory before the Hermes job can overwrite it.

## Testing
Manually tested and relied on existing tests to ensure no regressions.

### DIFF
--- a/features/fixtures/rn060/android/app/build.gradle
+++ b/features/fixtures/rn060/android/app/build.gradle
@@ -77,7 +77,7 @@ import com.android.build.OutputFile
 
 project.ext.react = [
     entryFile: "index.js",
-    enableHermes: System.env.RN_ENABLE_HERMES == true,  // clean and rebuild if changing
+    enableHermes: !!System.env.RN_ENABLE_HERMES // clean and rebuild if changing
 ]
 
 apply from: "../../node_modules/react-native/react.gradle"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## TBD
+
+* Upload correct bundle when Hermes is enabled
+  [404](https://github.com/bugsnag/bugsnag-android-gradle-plugin/pull/404)
+
 ## 5.7.7 (2021-06-23)
 
 * Emit a helpful compatibility error when Android Gradle Plugin 7 or higher is detected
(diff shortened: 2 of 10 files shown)


features/fixtures/rn060/android/app/build.gradle
```diff
@@ -77,7 +77,7 @@ import com.android.build.OutputFile
 
 project.ext.react = [
     entryFile: "index.js",
-    enableHermes: System.env.RN_ENABLE_HERMES == true,  // clean and rebuild if changing
+    enableHermes: !!System.env.RN_ENABLE_HERMES // clean and rebuild if changing
```
fractalwrench, line 80: Does this handle when `System.env.RN_ENABLE_HERMES` is `null`?

src/main/kotlin/com/bugsnag/android/gradle/internal/ReactNativeUtils.kt
```diff
@@ -0,0 +1,45 @@
+package com.bugsnag.android.gradle.internal
+
+import com.android.build.gradle.api.ApkVariantOutput
+import org.gradle.api.Action
+import org.gradle.api.Describable
+import org.gradle.api.Project
+import org.gradle.api.Task
+import java.io.File
+
```
fractalwrench, line 9: nit: I feel like this might fit better in `BugsnagUploadJsSourceMapTask` as then all the sourcemap code remains in one place. I'm in two minds so will leave the decision up to you

fractalwrench (approved):
LGTM pending CI and resolution of comments

features/fixtures/rn060/android/app/build.gradle
lemnik: Yes, `null` will be `false` (as will an empty string).

PR #405, opened 2021-07-06
Update Maze-runner to version 5

Cawllec:
## Goal

- Updates Maze-runner to version 5, ensuring it can accept have any bug fixes or new features added when released, as well as being consistent with other repos
- Tidies up some of the request and environment logic
- Moves away from using unspecified payload queue steps to using type specific payload queue steps (requires https://github.com/bugsnag/maze-runner/pull/261 to be released)

## Todo:
- https://github.com/bugsnag/maze-runner/pull/261 needs to be released in next minor maze-runner version, with Gemfile moving back to targeting that release instead of a specific branch.

### DIFF
--- a/features/default_app.feature
+++ b/features/default_app.feature
@@ -2,7 +2,7 @@ Feature: Plugin integrated in default app
 
 Scenario: Single-module default app builds successfully
     When I build "default_app" using the "standard" bugsnag config
-    And I wait to receive 2 requests
+    And I wait to receive 2 builds
 
     Then 1 requests are valid for the build API and match the following:
       | appVersionCode | appVersion | buildTool      | sourceControl.provider | sourceControl.repository                                     |
--- a/features/react_native.feature
+++ b/features/react_native.feature
@@ -4,7 +4,7 @@ Feature: Plugin integrated in React Native app
 @skip_gradle_7_or_higher
 Scenario: Source maps are uploaded when assembling an app with the default project structure
     When I build the React Native app
-    And I wait to receive 3 requests
+    And I wait to receive 3 builds
 
     Then 1 requests are valid for the build API and match the following:
       | appVersionCode | appVersion | buildTool      |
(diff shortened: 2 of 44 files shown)


features/default_app.feature
```diff
@@ -2,7 +2,7 @@ Feature: Plugin integrated in default app
 
 Scenario: Single-module default app builds successfully
     When I build "default_app" using the "standard" bugsnag config
-    And I wait to receive 2 requests
+    And I wait to receive 2 builds
```
fractalwrench, line 5: I'm not convinced that this is the right abstraction because there are two separate requests types here: requests sent to the Build API and requests sent to the Upload API. Is there any possibility that these could be split out in mazerunner, assuming it wouldn't take massive amounts of effort?

The important thing is getting everything on v5 so if it does require lots of effort then it'd probably be ok to skip this

features/react_native.feature
```diff
@@ -126,22 +126,21 @@ Scenario: Source maps are uploaded in an app using Hermes
...
-    And I wait to receive 5 requests
+    And I wait to receive 5 builds
     And 5 requests are valid for the JS source map API and match the following:
         | appVersionCode | appVersion | overwrite | dev   |
         | 5              | 2.45.beta  | false     | false |
         | 5              | 2.45.beta  | false     | false |
         | 5              | 2.45.beta  | false     | false |
         | 5              | 2.45.beta  | false     | false |
         | 5              | 2.45.beta  | false     | false |
-    And the exit code equals 0
```
fractalwrench, line 137: Is there a reason for the removal of this assertion?

fractalwrench (commented):
Had a couple of queries which I've left inline. Also, it's worth noting that once this is merged it'll probably also need to go into the `v7/master` branch where development is happening for BAGP v7.

features/react_native.feature
Cawllec: Yes, I was going to make it implicit to running the react-native and NDK scripts, which I hadn't pushed up.  It's now been added to the branch.

features/default_app.feature
twometresteve: It's not a lot of effort - I'll add another "uploads" endpoint to MazeRunner now, which should also be of use for other tools in time.

twometresteve: Addition of the new endpoint in Maze Runner is now in development and I'm updating the BAGP tests against the development branch.  Once it's all lined up, I'll create a new PR into this one that updates all the affected scenarios (there are a lot!).

fractalwrench (approved):

PR #408, opened 2021-07-16
Address task dependency warning when using APK splits

fractalwrench:
## Goal

Note: this is a backport of #407.

Addresses a warning about implicit [task dependencies](https://docs.gradle.org/7.0.2/userguide/validation_problems.html#implicit_dependency) that is displayed when using Gradle 7 with APK splits. This is caused by the outputs of tasks that generate mapping files being the same directory for several tasks - the fix has been to alter this so the directory is based on the variant output's name.

## Testing

Verified the message shows in an example app using AGP 7 with splits, then tested with the local changes and confirmed that the error message is no longer displayed. Otherwise relied on existing test coverage.


lemnik (approved):
LGTM

PR #409, opened 2021-07-21
Add clearer error messages if the node_modules directory is not found

lemnik:
## Goal
Emit a clear error message when the node_modules directory cannot be located by the plugin, or if the Bugsnag react-native module cannot be found. Previously the plugin would create an invalid Maven reference, potentially leading to compiler errors.

## Testing
Manually tested by removing the `node_modules` directory and running the Gradle build

### DIFF
--- a/src/main/kotlin/com/bugsnag/android/gradle/BugsnagPlugin.kt
+++ b/src/main/kotlin/com/bugsnag/android/gradle/BugsnagPlugin.kt
@@ -340,9 +340,25 @@ class BugsnagPlugin : Plugin<Project> {
             project.rootProject.allprojects { subProj ->
                 val defaultNodeModulesDir = File("${subProj.rootDir}/../node_modules")
                 val nodeModulesDir = bugsnag.nodeModulesDir.getOrElse(defaultNodeModulesDir)
+                if (!nodeModulesDir.exists()) {
+                    throw StopExecutionException(
+                        "Cannot find node_modules directory at: ${nodeModulesDir.absolutePath} " +
+                            "To set this to the correct path manually, please see: " +
+                            "https://docs.bugsnag.com/build-integrations/gradle/#custom-node_modules-directory"
+                    )
+                }
+
+                val bugsnagModuleDir = File(nodeModulesDir, "@bugsnag/react-native/android")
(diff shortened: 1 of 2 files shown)


src/main/kotlin/com/bugsnag/android/gradle/BugsnagPlugin.kt
```diff
@@ -340,9 +340,25 @@ class BugsnagPlugin : Plugin<Project> {
...
+                            "To set this to the correct path manually, please see: " +
+                            "https://docs.bugsnag.com/build-integrations/gradle/#custom-node_modules-directory"
+                    )
+                }
+
+                val bugsnagModuleDir = File(nodeModulesDir, "@bugsnag/react-native/android")
+                if (!bugsnagModuleDir.exists()) {
+                    throw StopExecutionException(
+                        "Cannot find the @bugsnag/react-native module in your node_modules directory. " +
+                            "Manual installation instruction can be found here: " +
```
twometresteve, line 355:
```suggestion
                            "Manual installation instructions can be found here: " +
```

?

src/main/kotlin/com/bugsnag/android/gradle/BugsnagPlugin.kt
```diff
@@ -340,9 +340,25 @@ class BugsnagPlugin : Plugin<Project> {
             project.rootProject.allprojects { subProj ->
                 val defaultNodeModulesDir = File("${subProj.rootDir}/../node_modules")
                 val nodeModulesDir = bugsnag.nodeModulesDir.getOrElse(defaultNodeModulesDir)
+                if (!nodeModulesDir.exists()) {
+                    throw StopExecutionException(
+                        "Cannot find node_modules directory at: ${nodeModulesDir.absolutePath}. " +
```
xander-jones, line 345:
```suggestion
                        "Cannot find node_modules directory at: ${nodeModulesDir.absolutePath}" +
```

Wondering if that trailing dot might cause unnecessary confusion?

xander-jones (approved):

fractalwrench (approved):
LGTM once other comments are resolved - this should make the error messaging much nicer.

It'll also need targeting at the v7 branch once CI has finished doing its thing.

PR #410, opened 2021-07-22
Add clearer error messages if the node_modules directory is not found

lemnik:
## Goal
Emit a clear error message when the node_modules directory cannot be located by the plugin, or if the Bugsnag react-native module cannot be found. Previously the plugin would create an invalid Maven reference, potentially leading to compiler errors.

## Testing
Manually tested by removing the `node_modules` directory and running the Gradle build


fractalwrench (approved):
LGTM

PR #414, opened 2021-08-05
task(tests): add unity 2021 gradle export as fixture

rich-bugsnag:
## Goal

Create a Unity 2021 testing fixture

## Changeset

Added a gradle export in fixtures/unity_2021


fractalwrench (approved):
LGTM

PR #416, opened 2021-08-09
Access version code lazily via provider

### LINKED ISSUE #384: versionCode is always resolved at configuration time
### Describe the bug
the `versionCode` property is always resolved at configuration time. This is expensive as it could very well be computed by a task, meaning this is now executing a task at configuration time

https://github.com/bugsnag/bugsnag-android-gradle-plugin/blob/master/src/main/kotlin/com/bugsnag/android/gradle/BugsnagPlugin.kt#L649

It should try to use the newer property-based API output API to avoid this on newer AGP versions where possible

### Steps to reproduce
All the details are above. Reproduction is easy - set a custom version code mapped from a gradle property. In our case we do it from an env var on CI.

### Environment
* Android Studio version: N/A
* Android Gradle version: 7
* Bugsnag manifest section (if modified): N/A

jrodbx:
Closes #384.

## Goal

To unblock consumers (such as Square/CashApp) to use the Bugsnag Android plugin with AGP 7.0 final.

## Design

The inevitable fix would be to switch from using the now-legacy variant api:
```
            val android = project.extensions.getByType(AppExtension::class.java)
            android.applicationVariants.configureEach { variant ->
```

to the new provider api, a la:
```
            val android = project.extensions.getByType(ApplicationAndroidComponentsExtension::class.java)
            android.onVariants { variant -> 
                variant.outputs.map { output -> output.versionCode }
            }
```

but that would have been a more substantial change.

## Changeset

Wrapping output.versionCode in a Callable which can be passed to a DefaultProvider for lazy initialization.

## Testing

I pushed to our internal Artifactory instance and ran the following command successfully:

```
./gradlew publishInternalReleaseBundle --dry-run
```

as outlined in https://github.com/Triple-T/gradle-play-publisher/issues/940#issue-862925799

cc: @egorand @SUPERCILEX

### DIFF
--- a/src/main/kotlin/com/bugsnag/android/gradle/BugsnagPlugin.kt
+++ b/src/main/kotlin/com/bugsnag/android/gradle/BugsnagPlugin.kt
@@ -187,7 +187,7 @@ class BugsnagPlugin : Plugin<Project> {
             val manifestInfoFileProvider = registerManifestUuidTask(project, variant)
             val mappingFilesProvider = createMappingFileProvider(project, variant, output)
 
-            val versionCodeProvider = DefaultProvider { output.versionCode }
+            val versionCodeProvider = project.providers.provider { output.versionCode }
 
             val generateProguardTaskProvider = when {
                 jvmMinificationEnabled -> registerGenerateProguardTask(
--- a/src/main/kotlin/com/bugsnag/android/gradle/BugsnagPlugin.kt
+++ b/src/main/kotlin/com/bugsnag/android/gradle/BugsnagPlugin.kt
@@ -38,6 +38,7 @@ import org.gradle.api.Task
 import org.gradle.api.UnknownTaskException
 import org.gradle.api.file.FileCollection
 import org.gradle.api.file.RegularFile
+import org.gradle.api.internal.provider.DefaultProvider
 import org.gradle.api.provider.Provider
 import org.gradle.api.tasks.StopExecutionException
 import org.gradle.api.tasks.TaskProvider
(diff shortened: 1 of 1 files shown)


src/main/kotlin/com/bugsnag/android/gradle/BugsnagPlugin.kt
```diff
@@ -186,11 +187,14 @@ class BugsnagPlugin : Plugin<Project> {
             val manifestInfoFileProvider = registerManifestUuidTask(project, variant)
             val mappingFilesProvider = createMappingFileProvider(project, variant, output)
 
+            val versionCodeProvider = DefaultProvider { output.versionCode }
```
ZacSweers, line 190: You should use providerfactory instead of internal APIs

```suggestion
            val versionCodeProvider = project.providers.provider { output.versionCode }
```

ZacSweers (approved):
lgtm sans the internal API usage but it would be better to just use the modern AGP APIs and pipe the versionCode property from the new variant API

src/main/kotlin/com/bugsnag/android/gradle/BugsnagPlugin.kt
```diff
@@ -38,6 +38,7 @@ import org.gradle.api.Task
 import org.gradle.api.UnknownTaskException
 import org.gradle.api.file.FileCollection
 import org.gradle.api.file.RegularFile
+import org.gradle.api.internal.provider.DefaultProvider
```
jrodbx, line 41:
```suggestion
```

yousif-bugsnag:
Hey @jrodbx, thanks for the PR! We'll look to review this as soon as priorities allow.

lemnik:
Hi @jrodbx / @ZacSweers.

Thanks for this PR and your input. I hope you don't mind that we took this as inspiration for the logic here: https://github.com/bugsnag/bugsnag-android-gradle-plugin/pull/427/files#diff-0b3ba5e394543c53bbd77a52b87f07c4ff12e47518f2f3b26f1cd28f8246d7fcR218

This allows us to cover the entire Manifest merge including `versionCode` and `versionName` overrides in variants, along with handling our own `meta-data` overrides that can appear in `AndroidManifest`. I believe the change solves the problems you were having, and was released as part of v7.1.0.

Please let us know if this is still an issue - and thanks again.

PR #418, opened 2021-08-16
task(unity): recreate unity 2021 export with il2cpp

rich-bugsnag:
The previous export was mono and @lemnik requested it be changed to IL2CPP


rich-bugsnag:
No longer required

PR #419, opened 2021-08-17
Add support for the Unit 2021 symbols.zip files

lemnik:
## Goal
Support Unity 2021 format `symbols.zip` files, where the shared-objects no longer include the `.sym.so` filename pattern.

## Design
If a discovered shared-object file (`symbols.zip`) does not contain any `.sym.so` files it is assumed to be a Unity 2021 file. If this is the case we extract `libil2cpp.so`, `libmain.so`, and `libunity.so` and then continue with the existing process.

## Testing
New unity test scenario and fixture


kattrali (approved):

rich-bugsnag (approved):
LGTM, let me know if you want me to test within a unity project 👍

PR #423, opened 2021-09-02
Support extracting SO files from bugsnag AARs for library modules

fractalwrench:
## Goal

Supports extracting SO files from bugsnag AARs for library modules. This allows linking against bugsnag's NDK API without requiring users to write their own gradle task.

## Changeset

- Updated usage of `ApkVariant` and `ApkVariantOutput` to the parent interfaces `BaseVariant` and `BaseVariantOutput`. This will make it easier in future to support uploading from library modules if we decide to do so
- Logged a warning that library modules do not support uploading mapping files
- Registered the `BugsnagInstallJniLibsTask` when in a library module

## Testing

- Created E2E test fixtures that are used to build an app module and a library module which link against bugsnag-android's AARs.
- Altered E2E tests to verify the expected behaviour.
- Updated the default AGP version used to run tests locally to 7.0.0


lemnik (approved):
LGTM

PR #425, opened 2021-09-06
Reject versionCodeOverride being used with VERSION_CODE meta-data

lemnik:
## Goal
The Android Gradle Plugin allows each application variant to have it's own `versionCode` and `versionName`, while Bugsnag also allows `versionCode` overriding using an `AndroidManifest.xml` meta-data attribute (`com.bugsnag.android.VERSION_CODE`). Using these two attributes together causes unexpected behaviour, with the source-maps being uploaded with one version and reports generated with a different version.

When detecting both a `versionCodeOverride` and `com.bugsnag.android.VERSION_CODE` in the same project we will now emit an error explaining that they should not be used together.

## Design
The default `versionCode` / `versionName` attributes are tracked separately to those provided in `meta-data` attributes, and the final version details for a variant output are determined by a new `AndroidManifestInfo.forApkVariantOutput` function. This logic has been lifted into the `BugsnagPlugin` and the tasks requiring the version details receive their final `AndroidManifestInfo` by way of a mapped `Provider`.

## Testing
New unit tests and end-to-end tests.

### DIFF
--- a/src/main/kotlin/com/bugsnag/android/gradle/BugsnagReleasesTask.kt
+++ b/src/main/kotlin/com/bugsnag/android/gradle/BugsnagReleasesTask.kt
@@ -22,13 +22,10 @@ import org.gradle.api.provider.MapProperty
 import org.gradle.api.provider.Property
 import org.gradle.api.provider.ProviderFactory
 import org.gradle.api.tasks.Input
-import org.gradle.api.tasks.InputFile
 import org.gradle.api.tasks.InputFiles
 import org.gradle.api.tasks.Internal
 import org.gradle.api.tasks.Optional
 import org.gradle.api.tasks.OutputFile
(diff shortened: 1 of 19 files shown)


src/main/kotlin/com/bugsnag/android/gradle/BugsnagReleasesTask.kt
```diff
@@ -60,20 +57,15 @@ sealed class BugsnagReleasesTask(
         description = "Assembles information about the build that will be sent to the releases API"
     }
 
+    @get:Input
+    override val manifestInfoFile: Property<AndroidManifestInfo> = objects.property()
```
fractalwrench, line 61: nit: might be worth renaming the property names across these tasks as this is no longer a File

fractalwrench (approved):

PR #426, opened 2021-09-07
Support DexGuard with AGP 7

fractalwrench:
## Goal

Supports the DexGuard plugin with AGP 7. This avoids eager configuration of the DexGuard task that the Bugsnag plugin needs to rely upon, which was triggering an exception in projects using DexGuard 9.1.12.

## Testing

Manually verified that an example app using DexGuard can upload mapping information to Bugsnag when assembling + bundling.


lemnik (approved):
LGTM

PR #432, opened 2021-09-15
Reinstate React Native E2E tests for AGP 4.2

fractalwrench:
## Goal

Enables the React Native E2E tests for AGP 4.2 by using Gradle 6.8. This circumvents a bug in the RN CLI that prevents use of older RN versions with Gradle 7.

The fixture in v7 will be updated to use Gradle 7 + RN 0.65.


lemnik (approved):
LGTM

PR #441, opened 2021-09-21
Fix task ordering for AGP 3

lemnik:
## Goal
Fix projects using Android Gradle Plugin 3.5 with Proguard by ensuring that our generation and upload tasks run after the main build process has been completed.

## Design
Due to how the Proguard `mapping.txt` is handled in AGP < 4 we need to manually enforce a task ordering to ensure that our Proguard generation and upload tasks run after the `mapping.txt` is actually generated.

## Testing
ReactNative Monorepo fixture configured for `enableProguardInReleaseBuilds`. Manual testing with various versions of AGP and ReactNative projects.


fractalwrench (approved):
LGTM pending CI

PR #451, opened 2022-01-05
Fix FileAlreadyExistsException when using ReactNative & Hermes

lemnik:
## Goal
Fix [#1550](https://github.com/bugsnag/bugsnag-js/issues/1550) by ensuring that existing source-maps are overwritten instead of causing build failures.

## Testing
Tested manually by reproducing the error as described, and then testing that the fix no-longer causes the same issue. Instead the source maps are now overwritten and the new source maps are uploaded.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## TBD
+
+# Bug Fixes
+
+* Fix FileAlreadyExistsException errors when building ReactNative projects with Hermes
+  [#451](https://github.com/bugsnag/bugsnag-android-gradle-plugin/pull/451)
+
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+## TBD
+
+# Bug Fixes
+
+* Fix FileAlreadyExistsException errors when building ReactNative projects with Hermes
+  []()
```
fractalwrench, line 6: This is missing a link

fractalwrench (approved):
LGTM pending resolution of missing link in changelog

Issue #444, opened 2021-09-22
Incompatibility between Bugsnag Gradle Plugin 7.1.0 and DexGuard 1.1.14/1.1.15


guilherme-pereira-ifood:
### Describe the bug
We were using Bugsnag Gradle Plugin 7.0.0 + DexGuard Gradle Plugin 1.1.14 and we could build the project just fine. We're trying to update Bugsnag Gradle Plugin to 7.1.0 but it fails the build on the configuration phase.

We tried to update the DexGuard to the 1.1.15 (which was released today) and it failed with the same error.

This is the build output:

```
* What went wrong:
A problem occurred configuring project ':app'.
> Task with name 'dexguardAabColombiaRelease' not found in project ':app'.

...

at com.bugsnag.android.gradle.BugsnagPlugin.registerBugsnagTasksForVariant(BugsnagPlugin.kt:205)
	at com.bugsnag.android.gradle.BugsnagPlugin.access$registerBugsnagTasksForVariant(BugsnagPlugin.kt:67)
	at com.bugsnag.android.gradle.BugsnagPlugin$setupBugsnagPlugin$1$1.execute(BugsnagPlugin.kt:157)
```

### Steps to reproduce
You'll need a project with build variants + build flavors.

I'll try to build a sample and attach it here.

### Environment
* Gradle version: 7.2
* Android Gradle Plugin (AGP) version: 7.0.2
* Bugsnag Android Gradle Plugin version: 7.1.0
* DexGuard Gradle Plugin version: 1.1.14 and 1.1.15

yousif-bugsnag:
Hi @guilherme-pereira-ifood, I've tried building a project with build variants and flavors, using those versions (Gradle 7.2, AGP 7.0.2, BAGP 7.1.0, DGP 1.1.15), but wasn't able to reproduce this. If you are able to provide a cut down reproduction case that demonstrates the issue that would be really useful.

guilherme-pereira-ifood:
Hi @yousif-bugsnag.

I built a reproducible in this repository: https://github.com/guilherme-pereira-ifood/DexGuardAGP7Issue

This repository is using bugsnag gradle plugin 7.0.0. Just change it to 7.1.0 on root `build.gradle` and sync the project. You'll be able to see the error happening.

Please check `app/build.gradle` to understand what we may be doing differently from your tests.

guilherme-pereira-ifood:
Hi @yousif-bugsnag.

[...]

PR #452, opened 2022-01-06
Fix "Task not found" failures when using DexGuard

lemnik:
## Goal
Fix [#444](https://github.com/bugsnag/bugsnag-android-gradle-plugin/issues/444) where builds with custom variants and DexGuard could fail with a "Task not found" error where limited build flavours where configured for DexGuard.

## Design
Check for variant level DexGuard configurations and if they exist: check each variant is configured in DexGuard before attempting to create an upload task for that variant.

## Testing
Manually tested with a known failing project.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@
 * Address task dependency warning when using APK splits
   [#412](https://github.com/bugsnag/bugsnag-android-gradle-plugin/pull/412)
 
+* Fix "Task with name *** not found" errors when using custom build variants and DexGuard
+  [#452](https://github.com/bugsnag/bugsnag-android-gradle-plugin/pull/452)
+
 ## 7.1.0 (2021-09-20)
 
 * Custom Dexguard paths are resolved correctly (relative to the project)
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,9 @@
 * Address task dependency warning when using APK splits
   [#412](https://github.com/bugsnag/bugsnag-android-gradle-plugin/pull/412)
 
+* Fix "Task with name *** not found" errors when using custom build variants and DexGuard
+  []()
```
fractalwrench, line 10: This is missing a link

src/main/kotlin/com/bugsnag/android/gradle/BugsnagPlugin.kt
```diff
@@ -346,11 +347,9 @@ class BugsnagPlugin : Plugin<Project> {
                 // DexGuard 9 runs as part of the bundle task supplied by AGP,
                 // so need to alter task dependency so that BAGP always runs
                 // after DexGuard
-                if (project.hasDexguardPlugin()) {
+                if (project.hasDexguardPlugin() && project.isDexguardEnabledForVariant(variant)) {
                     val taskName = getDexguardAabTaskName(variant)
-                    project.tasks.named(taskName).configure { dexguardTask ->
```
fractalwrench, line 351: It'd make sense to keep this configure block as otherwise I think this change could trigger eager configuration of the task

fractalwrench (approved):
LGTM pending review comments + manual testing

src/main/kotlin/com/bugsnag/android/gradle/BugsnagPlugin.kt
lemnik: The issue is that makes the order of `apply plugin` significant - the Dexguard plugin must be applied before BAGP otherwise the Dexguard task doesn't exist when we look it up here - also resulting in a `Task not found` exception.

By specifying the task by name rather than by reference we only resolve it at the end of configuration.

lemnik: Resolved this much more correctly by postponing the Dexguard task lookup using `generateProguardTaskProvider.configure`.

PR #464, opened 2022-05-19
AGP 7.2 support

lemnik:
## Goal
Basic support for Android Gradle Plugin 7.2.

## Design
`ExternalNativeBuildTask.soFolder` and `objFolder` return `File` in AGP 7.0 and `DirectoryProperty` in 7.2. To work around this we now use reflection on these properties and handle either case based on the returned value.

This PR also includes a change to the test fixture, using `sed` to alter the `gradle-wrapper.properties` avoiding versioning problems with the `gradle wrapper` task (which can have conflicts between the Gradle version and AGP version intended to be tested).

## Testing
New test pipeline for AGP 7.2 using existing fixtures.

### DIFF
--- a/features/scripts/setup_gradle_wrapper.sh
+++ b/features/scripts/setup_gradle_wrapper.sh
@@ -3,42 +3,34 @@ set -ex
 
 export UPDATING_GRADLEW=true
 
-pushd "$APP_FIXTURE_DIR"
-echo "Updating gradle wrapper for: $APP_FIXTURE_DIR"
-./gradlew wrapper --gradle-version "$GRADLE_WRAPPER_VERSION"
-popd
+update_gradle_wrapper_properties () {
+    sed -i.bak -e "s/^distributionUrl.*$/distributionUrl=https\\\\:\/\/services.gradle.org\/distributions\/gradle-$GRADLE_WRAPPER_VERSION-bin.zip/" gradle/wrapper/gradle-wrapper.properties
+}
 
-pushd "$NDK_FIXTURE_DIR"
-echo "Updating gradle wrapper for: $NDK_FIXTURE_DIR"
-./gradlew wrapper --gradle-version "$GRADLE_WRAPPER_VERSION"
-popd
+update_gradle_wrapper_in () {
+    pushd "$1"
+    echo "Updating gradle wrapper for: $1"
+    update_gradle_wrapper_properties
+    popd
+}
+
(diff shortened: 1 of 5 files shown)


features/scripts/setup_gradle_wrapper.sh
```diff
@@ -5,40 +5,40 @@ export UPDATING_GRADLEW=true
 
 pushd "$APP_FIXTURE_DIR"
 echo "Updating gradle wrapper for: $APP_FIXTURE_DIR"
-./gradlew wrapper --gradle-version "$GRADLE_WRAPPER_VERSION"
+sed -i.bak -e "s/^distributionUrl.*$/distributionUrl=https\\\\:\/\/services.gradle.org\/distributions\/gradle-$GRADLE_WRAPPER_VERSION-bin.zip/" gradle/wrapper/gradle-wrapper.properties
```
kattrali, line 8: Seems worthwhile to put this in a tiny function at the top of the file, since its the same somewhat-complex bits repeated a number of times.

src/main/kotlin/com/bugsnag/android/gradle/internal/ExternalNativeBuildTaskUtil.kt
```diff
@@ -0,0 +1,29 @@
...
+import org.gradle.api.file.DirectoryProperty
+import org.gradle.api.provider.Provider
+import org.gradle.api.provider.ProviderFactory
+import java.io.File
+import kotlin.reflect.full.memberProperties
+
+class ExternalNativeBuildTaskUtil(private val providerFactory: ProviderFactory) {
+    private fun getSearchDir(buildTask: Provider<ExternalNativeBuildTask>, propName: String): Provider<File> =
+        buildTask.flatMap { task ->
+            val soFolder = ExternalNativeBuildTask::class.memberProperties.find { it.name == propName }?.get(task)!!
```
kattrali, line 13: Does the unwrapped result here (`!!`) represent failure cases that can happen / should be handled?

lemnik: Not currently, both variants of the property are non-nullable in Kotlin.

features/scripts/setup_gradle_wrapper.sh
kattrali: :man_cook: :ok_hand:

kattrali (approved):

PR #469, opened 2022-07-08
Simplify Variant / VariantOutput override handling

lemnik:
## Goal
Work towards supporting Gradles configuration cache by ensuring all task Inputs are supported by the cache.

## Design
Moved the handling of `VariantOutput` related overrides (`versionCode` & `versionName`) into the `BugsnagManifestUuidTask` which adds the build UUID to the final `AndroidManifest.xml` file, and produces our internal `AndroidManifestInfo` file. This means that we now have an `AndroidManifestInfo` for each variant output, allowing each task (and therefore mapping file) to be matched to a single `AndroidManifestInfo` with the final data populated.

This unfortunately looses us the ability to detect conflicts between `versionCode` / `versionName` overrides, and the Bugsnag `com.bugsnag.android.VERSION_CODE` & `com.bugsnag.android.APP_VERSION` meta-data parameters. Instead the plugin will now upload with the same values as the app will use when reporting instead of emitting an error.

## Testing
Relied on our existing tests, while manually testing with configuration caching enabled.


src/main/kotlin/com/bugsnag/android/gradle/BugsnagManifestUuidTask.kt
```diff
@@ -1,24 +1,17 @@
...
+import org.gradle.api.tasks.Optional
 import org.gradle.api.tasks.OutputFile
+import org.gradle.api.tasks.PathSensitive
+import org.gradle.api.tasks.PathSensitivity
 import org.gradle.api.tasks.TaskAction
-import java.io.File
-import java.nio.file.Paths
 import javax.inject.Inject
 
 abstract class BaseBugsnagManifestUuidTask(objects: ObjectFactory) : DefaultTask() {
```
kattrali, line 17: Is the separate base class still needed?

lemnik: Nope. I'll remove it as well.

kattrali (approved):

PR #478, opened 2022-09-22
Fix search path for ndkBuild

lemnik:
## Goal
Avoid generating empty mapping files (or no mapping file) when `ndkBuild` is used.

## Changelog
`ndkBuild` generates an additional directory level (typically named `local`) between the expected `obj` directory and the architecture directory. This change looks for the `cxx` directory two or three levels above the directory reported from `ExternalNativeBuildTask`.

## Testing
Added support for `ndkBuild` to the existing `ndk` test fixture, and a test scenario to go with it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## TBD
+
+* Fixed a bug where using ndkBuild generated empty some mapping files which could not be uploaded
+  [#478](https://github.com/bugsnag/bugsnag-android-gradle-plugin/pull/478)
+
 ## 7.3.0 (2022-07-27)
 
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## TBD
+
+* Fixed a bug where using ndkBuild generated empty some mapping files which could not be uploaded
+  []()
```
kattrali, line 4:
```suggestion
  [#478](https://github.com/bugsnag/bugsnag-android-gradle-plugin/pull/478)
```

kattrali (approved):

PR #483, opened 2022-11-08
Merge NDK symbol upload tasks

lemnik:
## Goal

Merge the two NDK symbol upload tasks so that they don't need to be configured separately. This will avoid situations where `BugsnagPlugin` causes the configuration of AGP earlier than expected when detecting the appropriate mapping tool.

## Testing
Manual testing with a problematic project, along with the existing tests still passing.


src/main/kotlin/com/bugsnag/android/gradle/BugsnagUploadSoSymTask.kt
```diff
@@ -96,13 +120,43 @@ abstract class BugsnagUploadSoSymTask : DefaultTask(), AndroidManifestInfoReceiv
...
+
     companion object {
         private const val ENDPOINT_SUFFIX = "/ndk-symbol"
 
         private const val VALID_SO_FILE_THRESHOLD = 1024
 
-        fun taskNameFor(variant: BaseVariantOutput) =
-            "uploadBugsnag${variant.baseName.capitalize()}Symbols"
+        fun taskNameFor(variant: BaseVariantOutput, uploadType: UploadType) =
+            "uploadBugsnag${uploadType.name.toLowerCase().capitalize()}${variant.baseName.capitalize()}Mapping"
```
kattrali, line 159: Is this a _breaking_ change in some way?

kattrali (approved):

PR #491, opened 2023-01-20
Fix NdkToolchain.baseDir configuration cache

ZacSweers:
This property should be a `DirectoryProperty` rather than `Property<File>` in order to be compatible with configuration cache. This also allows for it to be more chainable with providers of this from android components.

## Goal

See above

## Design

See above

## Changeset

See above

## Testing

Manually tested in our project via maven local.

### DIFF
--- a/src/main/kotlin/com/bugsnag/android/gradle/internal/NdkToolchain.kt
+++ b/src/main/kotlin/com/bugsnag/android/gradle/internal/NdkToolchain.kt
@@ -8,20 +8,23 @@ import com.bugsnag.android.gradle.BugsnagGenerateUnitySoMappingTask
 import com.bugsnag.android.gradle.BugsnagPluginExtension
 import org.apache.tools.ant.taskdefs.condition.Os
 import org.gradle.api.Project
+import org.gradle.api.file.DirectoryProperty
 import org.gradle.api.logging.Logger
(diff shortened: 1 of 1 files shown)


ZacSweers:
Testing this reveals a separate issue though, which is that it's a required property for NdkToolchain and tasks to be configured, even if they aren't run. This presents a problem for consumers that don't have the NDK installed (even if they don't use it), as the property configuration then fails. There is a flag for this via the upload task control, but it's checked eagerly and hard to disable. I could take a crack at improving it here (i.e. make the baseDir property optional), or leave that for a followup.

johnkiely1:
Thanks for the PR @ZacSweers, we will review as soon as we can.

lemnik:
Hi @ZacSweers, thanks so much for this PR and pointing out the eager-config issue as well. Would you suggest making the `NdkToolchain` references a `Provider` instead of a hard-reference (specifically for tasks `register` functions)?

ZacSweers:
I'm not sure if that would help. I think where you'd want to end up is probably somewhere where you register the tasks regardless but make them no-op (± a warning log) or error if no NDK is found

lemnik:
@ZacSweers I've moved this commit over to: https://github.com/bugsnag/bugsnag-android-gradle-plugin/pull/499 so I'm closing this PR.

PR #496, opened 2023-02-10
Ensure Bugsnag mapping tasks run after Dexguard

lemnik:
## Goal
Ensure Bugsnag mapping tasks run after Dexguard so that the correct mapping files can be found and uploaded.

## Changelog
Corrected the task lookup to match the capitalisation used by Dexguard when no build flavour is being used.

## Testing
Tested locally

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## TBD
+
+### Bug Fixes
+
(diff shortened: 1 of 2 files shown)


tomlongridge (approved):
Needs a changelog though...

PR #499, opened 2023-02-14
Fix NdkToolchain.baseDir configuration cache

lemnik:
Internal PR for https://github.com/bugsnag/bugsnag-android-gradle-plugin/pull/491

## Goal
This property should be a DirectoryProperty rather than Property<File> in order to be compatible with configuration cache. This also allows for it to be more chainable with providers of this from android components.

## Testing
Relied on existing end-to-end testing


joshedney (approved):
lgtm!

PR #500, opened 2023-02-21
AGP-8 Compatibility

lemnik:
## Goal
Update the `bugsnag-android-gradle-plugin` to support Android Gradle Plugin 8.

## Changelog

- Update dependencies to support AGP-8 / Gradle 8
- Change the CI pipeline to use Java 14 (required by AGP-8)
- Remove the use of Gradle's deprecated `VersionNumber` class

## Testing
Modified existing test fixtures to be compatible with AGP-8.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -1,9 +1,9 @@
 steps:
-  - label: 'Audit current licenses'
-    timeout_in_minutes: 30
-    plugins:
-      - docker-compose#v3.7.0:
-          run: license-audit
+#  - label: 'Audit current licenses'
+#    timeout_in_minutes: 30
+#    plugins:
+#      - docker-compose#v3.7.0:
+#          run: license-audit
 
   - label: ':docker: Build Android gradle plugin CI image'
     key: 'agp-ci'
@@ -11,21 +11,21 @@ steps:
     plugins:
       - docker-compose#v3.7.0:
           build:
-            - android-gradle-plugin-ci
+            - android-gradle-plugin-ci-v8
           image-repository: 855461928731.dkr.ecr.us-west-1.amazonaws.com/android
           cache-from:
             - android-common:855461928731.dkr.ecr.us-west-1.amazonaws.com/android:latest
-            - android-gradle-plugin-ci:855461928731.dkr.ecr.us-west-1.amazonaws.com/android:android-gradle-plugin-ci-v7
+            - android-gradle-plugin-ci-v8:855461928731.dkr.ecr.us-west-1.amazonaws.com/android:android-gradle-plugin-ci-v8
       - docker-compose#v3.7.0:
           push:
-            - android-gradle-plugin-ci:855461928731.dkr.ecr.us-west-1.amazonaws.com/android:android-gradle-plugin-ci-v7
+            - android-gradle-plugin-ci-v8:855461928731.dkr.ecr.us-west-1.amazonaws.com/android:android-gradle-plugin-ci-v8
 
   - label: ':android: Detekt'
     depends_on: 'agp-ci'
     timeout_in_minutes: 30
     plugins:
       - docker-compose#v3.7.0:
-          run: android-gradle-plugin-ci
+          run: android-gradle-plugin-ci-v8
           command: ['./gradlew', 'detekt']
 
   - label: ':android: Ktlint'
@@ -33,7 +33,7 @@ steps:
     timeout_in_minutes: 30
     plugins:
       - docker-compose#v3.7.0:
-          run: android-gradle-plugin-ci
+          run: android-gradle-plugin-ci-v8
           command: ['./gradlew', 'ktlintCheck']
 
   - label: ':android: Unit tests'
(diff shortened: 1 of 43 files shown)


build.gradle
```diff
@@ -1,42 +1,20 @@
...
-    id "com.github.hierynomus.license" version "0.16.1"
-}
+    id("groovy")
 
-apply plugin: "org.jetbrains.kotlin.jvm"
-apply plugin: "org.jetbrains.kotlin.kapt"
-apply plugin: "io.gitlab.arturbosch.detekt"
-apply plugin: "org.jlleitschuh.gradle.ktlint"
+    id("org.jetbrains.kotlin.kapt") version "1.7.10"
+    id("org.jetbrains.kotlin.jvm") version "1.7.10"
```
tomlongridge, line 5: Should we use `kotlinVersion` here too? Or is it not necessarily the same thing?

src/main/kotlin/com/bugsnag/android/gradle/internal/AbstractSoMappingTask.kt
```diff
@@ -61,11 +61,7 @@ abstract class AbstractSoMappingTask : DefaultTask() {
         outputZipFile(process.inputStream, dst)
 
         val exitCode = process.waitFor()
-        if (exitCode != 0) {
-            throw IllegalStateException(
-                "Failed to generate symbols for $dst, objdump exited with code $exitCode"
-            )
-        }
+        check(exitCode == 0) { "Failed to generate symbols for $dst, objdump exited with code $exitCode" }
```
tomlongridge, line 64: Shouldn't this be `!=`?

src/main/kotlin/com/bugsnag/android/gradle/internal/ExternalNativeBuildTaskUtil.kt
```diff
@@ -1,31 +1,13 @@
...
         }
-
-    private fun fixNativeOutputPath(taskFolder: File): File {
-        return taskFolder.parentFile.parentFile.takeIf { it.parentFile.name == "cxx" }
-            ?: taskFolder.parentFile.parentFile.parentFile.takeIf { it.parentFile.name == "cxx" }
-            ?: taskFolder
-    }
-
-    fun findSearchPaths(buildTask: Provider<ExternalNativeBuildTask>) = arrayOf(
-        getSearchDir(buildTask, "objFolder"),
```
tomlongridge, line 28: Have we dropped support for this now?

src/main/kotlin/com/bugsnag/android/gradle/internal/NdkToolchain.kt
```diff
@@ -129,17 +129,17 @@ abstract class NdkToolchain {
          * Minimum `bugsnag-android` version where the new symbol uploading is available, using `objcopy` to produce
          * the symbol files instead of `objdump`
          */
-        internal val MIN_BUGSNAG_ANDROID_VERSION = VersionNumber.version(5, 26)
+        internal val MIN_BUGSNAG_ANDROID_VERSION = Version(5, 26, 0)
```
tomlongridge, line 132: nit: mix of styles to create a `Version` compared with internal objects above (e.g. `VERSION_6_6`) - parse vs constructor.

src/main/kotlin/com/bugsnag/android/gradle/internal/ExternalNativeBuildTaskUtil.kt
lemnik: Only the `soFolder` still exists, and the `getSearchDir` compatibility code is no longer required.

src/main/kotlin/com/bugsnag/android/gradle/internal/AbstractSoMappingTask.kt
lemnik: `check` is `if(!condition)` so it is equivalent to the previous code ("make sure exitCode == 0")

build.gradle
lemnik: We can't because the `plugins` block can't reference it directly. We should move to using a version catalogue, but I think that's beyond the scope of this PR?

.buildkite/pipeline.yml
```diff
@@ -1,93 +1,57 @@
...
-  - label: 'Audit current licenses'
-    timeout_in_minutes: 30
-    plugins:
-      - docker-compose#v3.7.0:
-          run: license-audit
+#  - label: 'Audit current licenses'
+#    timeout_in_minutes: 30
+#    plugins:
+#      - docker-compose#v3.7.0:
+#          run: license-audit
```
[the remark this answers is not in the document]
lemnik: 🤔  I took it out due to compatibility issues - but we should work to upgrade it as well.

src/main/kotlin/com/bugsnag/android/gradle/internal/AbstractSoMappingTask.kt
tomlongridge: Sorry - yes 🤦

.buildkite/pipeline.yml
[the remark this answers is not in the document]
tomlongridge: I've tasked this up separately.

tomlongridge (approved):

PR #514, opened 2023-04-12
NDK version detection based on package.xml

lemnik:
## Goal
Determine the NDK version being used based on the `package.xml` files within the NDK directories, with a fallback to the directory-name.

## Changelog

- Parse the `package.xml` in the NDK directory to retrieve the version number, allowing more flexible use of `ndkDir`
- Setting `useLegacyNdkSymbolUpload` now overrides the defaults inferred from the NDK version

## Testing
Manually tested with various NDK directories and options

PR #515, opened 2023-04-20
NDK version detection based on package.xml (v8)

lemnik:
## Goal
Determine the NDK version being used based on the `package.xml` files within the NDK directories, with a fallback to the directory-name.

## Changelog

- Parse the `package.xml` in the NDK directory to retrieve the version number, allowing more flexible use of `ndkDir`
- Setting `useLegacyNdkSymbolUpload` now overrides the defaults inferred from the NDK version

## Testing
Manually tested with various NDK directories and options

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## TBD
+
+### Enhancements
+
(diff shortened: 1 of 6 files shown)


tomlongridge (approved):
LGTM - compared with #514

PR #521, opened 2023-04-24
v8.0.0 - with enableNdkLinkage patch

lemnik:
v8.0.0 - With `enableNdkLinkage` patch

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,10 @@
-## 8.0.0 (2023-04-20)
+## 8.0.0 (2023-04-24)
 
 ### Enhancements
 
 * Determine the NDK version being used based on the `package.xml` files within the NDK directories, with a fallback to the directory-name.
   [#515](https://github.com/bugsnag/bugsnag-android-gradle-plugin/pull/515)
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,9 +1,11 @@
-## 8.0.0 (2023-04-20)
+## 8.0.0 (2023-04-24)
 
 ### Enhancements
 
 * Determine the NDK version being used based on the `package.xml` files within the NDK directories, with a fallback to the directory-name.
   [#515](https://github.com/bugsnag/bugsnag-android-gradle-plugin/pull/515)
+* `bugsnag.enableNdkLinkage` can be set to `false` to suppress the extraction of the `bugsnag-plugin-android-ndk` package in projects that don't require it, suppressing "Configuration was resolved at configuration time" warnings as a result.
```
tomlongridge, line 7: Would be good to include more on when the unpacking might be needed. Is this right?

```suggestion
* `bugsnag.enableNdkLinkage` can be set to `false` to suppress the extraction of the `bugsnag-plugin-android-ndk` package in projects that don't use [our NDK plugin](https://github.com/bugsnag/bugsnag-android/tree/master/bugsnag-plugin-android-ndk), suppressing "Configuration was resolved at configuration time" warnings as a result.
```

tomlongridge (approved):

PR #524, opened 2023-05-02
Remove external process execution from configuration phase

lemnik:
## Goal
Replace external process execution that happens during the configuration phase (used for `git --version`) with a `ValueHolder` to avoid breaking the configuration cache.

## Testing
Manually tested with Gradle 8.0 and 8.1

### DIFF
--- a/src/main/kotlin/com/bugsnag/android/gradle/BugsnagReleasesTask.kt
+++ b/src/main/kotlin/com/bugsnag/android/gradle/BugsnagReleasesTask.kt
@@ -1,6 +1,7 @@
 package com.bugsnag.android.gradle
 
 import com.bugsnag.android.gradle.internal.BugsnagHttpClientHelper
+import com.bugsnag.android.gradle.internal.GitVersionValueSource
 import com.bugsnag.android.gradle.internal.UploadRequestClient
 import com.bugsnag.android.gradle.internal.mapProperty
 import com.bugsnag.android.gradle.internal.property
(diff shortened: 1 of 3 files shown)


src/main/kotlin/com/bugsnag/android/gradle/BugsnagReleasesTask.kt
```diff
@@ -268,7 +269,8 @@ open class BugsnagReleasesTask @Inject constructor(
                 execSpec.standardOutput = baos
                 logging.captureStandardError(LogLevel.INFO)
             }
-            String(baos.toByteArray(), Charset.forName(CHARSET_UTF8)).trim { it <= ' ' }
+
+            baos.toString(Charset.defaultCharset()).trim { it <= ' ' }
```
lemnik, line 273: `ByteArrayOutputStream.toString(Charset)` is a Java 10 function that directly converts the internal byte array to a string instead of requiring a duplicate `byte[]`. AGP 8.+ require a minimum of Java 17

src/main/kotlin/com/bugsnag/android/gradle/internal/GitVersionValueSource.kt
```diff
@@ -0,0 +1,22 @@
...
+
+abstract class GitVersionValueSource : ValueSource<String, ValueSourceParameters.None> {
+    @get:Inject
+    abstract val execOperations: ExecOperations
+
+    override fun obtain(): String {
+        val output = ByteArrayOutputStream()
+        execOperations.exec {
+            it.commandLine("git", "--version")
+            it.standardOutput = output
```
tomlongridge, line 18: Does this handle the case when the user doesn't have `git` installed? Although I'm not entirely sure if the old way of calling it did either.

tomlongridge (approved):

PR #531, opened 2023-07-21
Add DexGuard version config option

SmartbearYing:
## Goal
Fix builds for users of DexGuard 9.4+

## Design
Due to the DexGuard 9.4+ Gradle plugin no longer having a VERSION property, default to DexGaurd 9 behaviour when DexGuard is being used.

Users of earlier versions can use:
```
bugsnag.dexguardMajorVersion = 8
```
to revert to DexGuard 8 compatible behaviour.

## Testing
Tested manually with DexGuard version 9.4 and DexGuard 8


lemnik (approved):
LGTM

PR #542, opened 2024-06-07
Fix documentation links in README.md

Egorand:
## Goal

It looks like the Gradle plugin documentation was moved from "gradle" to "android-gradle-plugin" folder, updated links accordingly and verified that they work.

## Design

N/A

## Changeset

Documentation links.

## Testing

N/A


clr182:
Hi @Egorand 

Thanks for sharing! 
We will look to merge your PR when priorities allow.

PR #544, opened 2024-11-28
change the default upload endpoint on v8

YYChen01988:
## Goal
Upload `mapping.txt` files to the `/proguard` endpoint when it is available.

## Testing

Existing tests

### DIFF
--- a/src/main/kotlin/com/bugsnag/android/gradle/BugsnagUploadProguardTask.kt
+++ b/src/main/kotlin/com/bugsnag/android/gradle/BugsnagUploadProguardTask.kt
@@ -82,7 +82,8 @@ open class BugsnagUploadProguardTask @Inject constructor(
         val manifestInfo = parseManifestInfo()
 
         // Send the request
-        val request = BugsnagMultiPartUploadRequest.from(this)
+        val newEndPoint = endpoint.get() + SUBDOMAIN_PROGUARD
+        val request = BugsnagMultiPartUploadRequest.from(this, newEndPoint)
         val mappingFileHash = mappingFile.md5HashCode()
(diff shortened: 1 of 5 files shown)


src/main/kotlin/com/bugsnag/android/gradle/BugsnagUploadProguardTask.kt
```diff
@@ -96,6 +97,7 @@ open class BugsnagUploadProguardTask @Inject constructor(
 
     companion object {
 
+        private const val SUBDOMAIN_PROGUARD = "/proguard"
```
lemnik, line 100: nit: formatting

```suggestion
    companion object {
        private const val SUBDOMAIN_PROGUARD = "/proguard"
        
```

src/main/kotlin/com/bugsnag/android/gradle/BugsnagUploadProguardTask.kt
```diff
@@ -82,7 +82,8 @@ open class BugsnagUploadProguardTask @Inject constructor(
         val manifestInfo = parseManifestInfo()
 
         // Send the request
-        val request = BugsnagMultiPartUploadRequest.from(this)
+        val newEndPoint = endpoint.get() + SUBDOMAIN_PROGUARD
```
lemnik, line 85: I think we should maybe ensure that the `endpoint` URL is in the expected format. Something like:

```suggestion
        val newEndPoint = endpoint.get().removeSuffix("/") + SUBDOMAIN_PROGUARD
```

Just in case the URL already ended with a `/`

lemnik (commented):
Our existing tests all specify an `endpoint` in their configuration, I think it's worth adding dedicated tests to ensure the new behavior is followed

lemnik (approved):

